PR #11, opened 2021-05-28
Updating areas with all of their climbs.

nadr0:
# Trello 

`OpenTacos development - nadr0` -> `leaf-area-page-md.js: list of climbs contained in this area`

# Task

Structuring and passing child climbs into the area/sector page. We want access to all of the `climbing-route` nodes that belong to the current area page you are on.

# Implementation

In `gatsby-node.js` I implement a trick to encode all of the `climbing-route` and `area-indices` with a `parentId` and `pathId`. 
> I didn't implement the `parentId` for the area files themselves because not every area has a parent `index.md` file in the folder.

The on the fly id system is implemented based on the folder path structure which we know will be unique. This allows us to render our tree structure without having to hardcode ids into the actual `.md` files. 

After passing the data to the sector page, I added an extra query to the page query for it will fetch all of the `climbing-routes` for the given area page. I projected some of the climb data out of the query.

Once I had all of the climbing data I looped through all of the nodes and rendered them in a grid view based on the original `netlify` example with my new `RouteCard.js` reusable component. This was just to show you that we can render the climbs on an area. 

Lastly, I added 2 more chip colors for `tr` and `boulder` for those can be visualized while implementing. 

# Example

http://localhost:8000/areas/105816095/iron-butte is a good example since it has a lot of climbs

image

# Next Steps

You mentioned in my other PR you would be open to change the layout of the page.

I don't mind the grid view and the verbose description view. I could continue to refactor that code to update the new sector page with this. Additionally you mentioned about rendering some of this with the `mdx` component? 

I think on the sector page it may make sense to render it with the React Component in the sector page. I don't know how it would look within the `mdx` renderer. I would have to learn how that works.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -111,20 +111,6 @@ exports.createPages = async ({ graphql, actions }) => {
       },
     });
   }
-  // result.data.allMdx.edges.forEach(({ node }) => {
-  //   const climbs = await retrieveClimbs(node.fields.pathId);
-  //   console.log(climbs);
-  //   createPage({
-  //     path: node.fields.slug,
-  //     component: path.resolve(`./src/templates/leaf-area-page-md.js`),
-  //     context: {
-  //       legacy_id: node.frontmatter.metadata.legacy_id,
-  //       // climbs: node.climbs,
-  //       // name: node.area_name,
-  //       slug: node.fields.slug,
-  //     },
-  //   });
-  // });
 
   // Query all route .md documents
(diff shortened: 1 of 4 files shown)


vnugent:
Nice work figuring out how to get all the climb nodes for an area!
You've got the mix-and-match of MDXRender and normal React rendering.  We'd use it for converting markdown text to JSX.  The rest can be rendered the normal react way.

PR #15, opened 2021-05-30
Nadro.area climbs ux windows support

nadr0:
#14  Is included in this PR. You should merge that one first.

# Trello

OpenTacos: Area explore view
OpenTacos development - nadr0

# Task

As a user I want to navigate an area and their climbs. 
As a user I want to be able to see my current path of which areas I am in. 

# Implementation 

- Implemented a BreadCrumbs component
- Implemented a AreaCard component, similar to a route card
- Implemented navigation from Area to child climb by clicking on card
- Implemented BreadCrumbs Navigation
- Rendered sub areas within an Area
- Implemented Area navigation by clicking on AreaCards
- Fixed the `pathId` and `parentId` system to use a POSIX format only. We just need a consistent path structure no matter what system someone is on. Since I realized the code would run differently on someone's machine. 
- Added dummy index.md data for sub areas to show you how the navigation works. 

# Example

Please use this URL to test locally http://localhost:8000/areas/d96ff3d2-d627-4fa7-8092-36baeb0d4ad6/western-nevada
These folders have the dummy index.md to create the navigation structure. 

image
image
image

# Next Steps

This should do the job of rendering all climbs and all sub areas given an area. It also will implement the navigation via RouteCards, AreaCards, and the BreadCrumbs component. 

The one downside of this implementation is every area needs an `index.md`.  Though I don't think this is really a downside.
I do not think we should have empty parent area folders. Based on our tree structure of the data every area should probably have an `index.md` file because anyone would be able to give each area a description.

If we migrate data into the content folder all areas should have an `index.md` file and if that is true then my rendering and navigation code above will work.

We should have these following definitions
- A climb is defined by a markdown file named <route-name>.md within an area.
- A area is defined by an index.md file within an area (recursive). 

i.g. You can't skip 15 empty folders down without creating all of the index.md for the previous areas. Those areas aren't just empty. Someone would be able to contribution to the local knowledge of that area in some way.


vnugent (approved):
Nice work with area breadcrumb navigation.  I agree with you that each area needs an index file.  The original dataset only has content for leaf areas.  I'll generate dummy files.

PR #20, opened 2021-06-04
Nadr0.ideas for main page 02

nadr0:
# Trello 

https://trello.com/c/uA1NGjSD/19-opentacos-ideas-for-the-main-page

# Task

On the main page I wanted to add a randomly featured route every time the user refreshes the page

# Implementation

- Added a new RandomRouteCard to render the format for the new randomly selected route
- Created a few reusable components for the grade chip as well as the climbing type chips
- I added a `useStaticQuery` to query all the climbing routes then select a random one from the list

This `useStaticQuery` option could be very slow if we are fetching a lot of data. It is quite quick for now. If this becomes too slow i'll have to remove it and think of a new way.

Currently there is no way to pass variables into `useStaticQuery`. The only way is on `page` renders but this is not a page that is created from a graphql query/gatsby-node.js. 

So for now I just fetch all the routes in the entire node tree then randomly select one, clear the unused array then pass the data into the card to render the random route.

I tried pass the `body` into the `useStaticQuery` and it was really really slow. So I know for a fact this can become an issue. That is why the card doesn't have a `body` value. 

# Example

image

# Next Steps

Figure out if there is a better way to randomly generate the route? Should we move this main page to a `createPage` function or we can pass a context into the query that renders the page?


src/pages/index.js
```diff
@@ -1,10 +1,52 @@
...
+              }
+            }
+          }
+        }
+    }
+  `);
+  const min = 0;
+  const max = allClimbingRoutes.allMdx.edges.length;
+  const randomIndex = getRandomInt(min,max);
+  const randomClimb = allClimbingRoutes.allMdx.edges[randomIndex].node;
```
vnugent, line 48: Once we have a larger dataset perhaps we could generate a list of say 20 random climbs at build time, and at runtime we cycle through that list.

vnugent (approved):

PR #54, opened 2021-08-15
Add an editor

vnugent:
### Trello: https://trello.com/c/Tk768QcH/39-opentacos-research-a-cms-editor

- Introduce a simple visual editor based on [Plate](https://plate.udecode.io/) 
- Plate provides a nice abstraction on top of Slate.js.  
I have a bit of experience working with Slate in the past – while it's a popular and powerful lib for building visual editors, it's still a low level library, and we would need to deal with the inner working of Slate, namely the document AST. 
- The edit component accessible by `/edit?file=<open-tacos content-repo/path to markdown file>.md` will load the raw markdown file from GitHub into the editor
- I think this PR contains enough logic for us to try to integrate with the Git gateway and/or protected route (auth0).  Clicking Submit will convert internal document ast to markdown string.

TODOs:
- [ ] Support frontmatter editing
- [ ] Support more formats such as H2, link, lists
- [ ] Enforce document schema, ie make the headers fixed, unremovable


src/components/editor/Editor.js
```diff
@@ -0,0 +1,44 @@
+import React, { useState, useEffect } from "react";
+import PlateEditor from "./PlateEditor";
+import axios from "axios";
+import queryString from "query-string";
+
+import fm from "front-matter";
+
+//TODO: make this a configurable option in gatsby-config.js
```
nadr0, line 8: Do we need to do this TODO within this PR or is it for later?

src/components/editor/components.js
```diff
@@ -0,0 +1,178 @@
...
+      //       cursor: pointer;
+      //       color: ${reversed
+      //         ? active
+      //           ? 'white'
+      //           : '#aaa'
+      //         : active
+      //         ? 'black'
+      //         : '#ccc'};
+      //     `
+      //   )}
```
nadr0, line 27: Do we need all of this?

src/components/editor/Editor.js
```diff
@@ -0,0 +1,44 @@
...
+export const Editor = () => {
+  const [value, setValue] = useState(null);
+  const [debug, setDebug] = useState(false);
+  useEffect(() => {
+    const get_file_from_github = async () => {
+      const parsed = queryString.parse(location.search);
+      if (parsed.file) {
+        const res = await client.get(
+          `${CONTENT_BRANCH}/content/${parsed.file}`
+        );
```
nadr0, line 20: If we are using await we should probably try catch this block and maybe show and error saying the page couldn't be fetched. 

i.g. if in the catch block we just return `<div>Something went wrong! Could not edit the page!</div>`

nadr0:
image


Verification that I was able to get it up and running locally. Awesome work.

src/components/editor/components.js
vnugent: removed emotion lib and dead code

nadr0 (approved):
Changes look great. Thanks for making those and implementing this feature!

PR #56, opened 2021-08-21
Nadro/auth0 with editor

nadr0:
# Important Notes

- All the configuration is setup to be hard coded to `https://localhost:8000` and `https://localhost:8000/edit` for callback URLs within the Auth0 application. So if this gets to deployed to dev or main it obviously won't work. I can help configure this later. 
- They said in their documentation `"With the exception of custom URI schemes for native clients, all callbacks should use protocol https://"`; This means that I had to run localhost with HTTPS. Assuming you have an SSL binary installed globally you can just run `npm run develop -- --https` and it will just work. 
- I need to refactor the client side configuration variables for auth0 into environment variables to share them and clean up the code. 
- Auth0 does not allow wildcard (`*`) all back URLs. This means we must save their current page and when they are redirected we auto redirect them back to the page they were on. It is weird but I don't know another way around this. 
- I added a dummy `callAPI` function that will get us our `accessToken` which we will want to use to authenticate the request of the user when they ping Git Gateway.
- I have not setup user or app metadata or roles for the accounts created. 
- Lastly, do we want 1 application of accounts for both dev and app? or two application Ids? 

# Task 

We want users to be able to login to Auth0 to have an account that will allow them to edit routes and area pages.

# Implementation

- I added the auth0 react NPM package which allows us to login, logout, and fetch the access token from the user.
- I implemented a protected route page under `/edit` which won't render the Editor.js unless they are logged in. 

# Testing

Going to an edit page without being logged in 
image

Auth0 prompt from the redirect
image

After you login
image


gatsby-node.js
```diff
@@ -235,13 +235,29 @@ exports.createPages = async ({ graphql, actions }) => {
...
  */
-exports.onCreateWebpackConfig = ({ actions }) => {
+ exports.onCreateWebpackConfig = ({ actions }) => {
   actions.setWebpackConfig({
     resolve: {
       fallback: {
         path: require.resolve("path-browserify"),
-        "assert": false
+        "assert": false,
+        "stream": false
```
nadr0, line 244: I had a webpack polyfill issue so I needed to add `"stream":false`

src/components/auth0NavBar/Auth0Profile.js
```diff
@@ -0,0 +1,42 @@
...
+import { useAuth0 } from "@auth0/auth0-react";
+
+const Auth0Profile = () => {
+  const { user, isAuthenticated, getAccessTokenSilently } = useAuth0();
+
+  /**
+   * Allows us to fetch the accessToken that we will need for git gateway.
+   */
+  const callApi = async () => {
+    const domain = "dev-fmjy7n5n.us.auth0.com";
```
nadr0, line 11: Domain that is automatically generated for my Single Page Application within Auth0.

src/components/auth0NavBar/LogoutButton.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { useAuth0 } from "@auth0/auth0-react";
+
+const LogoutButton = () => {
+  const { logout, isAuthenticated } = useAuth0();
+
+  return isAuthenticated && (
+    <button className="btn btn-candy" onClick={() => logout({ returnTo: window.location.origin })}>
```
nadr0, line 8: Logs you out to the home page since we cannot wildcard pages I just redirect them back to `https:localhost:8000/`

src/pages/edit.js
```diff
@@ -1,17 +1,43 @@
...
+  if (query.includes("file=")) {
+    window.localStorage.setItem('previousPath', window.location.href);
+  }
+
+  useEffect(async () => {
+    const query = window.location.search;
+    if (query.includes("code=") && query.includes("state=")) {      
+      const href = window.localStorage.getItem('previousPath');
+      window.location.href = href;
+    }
```
nadr0, line 20: Since we cannot wild card the login redirect 

i.g.  `/edit?file=1234` -> login -> login success -> `/edit` 

You cannot do `/edit*` in the Auth0 configuration. So I save off the previous path they were at in the local storage. When the user gets redirected back to `/edit` and it is a success they are then redirected back to their original page they were on since we read it from local storage. It is kinda a hack but that is what Auth0 recommended.

OR you could do this within a callback redirect flow by passing data through the login request then reading that value when they get redirected but they have absolutely zero documentation on this. They just mention it but don't provide code snippets.

src/pages/edit.js
```diff
@@ -1,17 +1,43 @@
...
+  });
+
   return (
+    <Auth0Provider
+    domain="dev-fmjy7n5n.us.auth0.com"
+    clientId="0Eth3vOcH6IY75szP00kVQmQt8KhzZmL"
+    redirectUri="https://localhost:8000/edit"
+    audience="https://dev-fmjy7n5n.us.auth0.com/api/v2/"
+    scope="read:current_user update:current_user_metadata"
```
nadr0, line 29: Should be configured within GATSBY_ environment variables that are client side exposed.

src/pages/edit.js
vnugent: - I pushed some changes to your branch as I was having issue with the edit page not loading the editor after logging in (tested with Brave and Chrome).  I ended up refactoring edit.js, moving Auth0 component to app's root component per their Gatsby example[1] for it to interact the url router.

- Protect the edit page with `export default withAuthenticationRequired(edit)`.  It takes care of the redirect after login.

[1] Auth0 Gatsby example:
https://github.com/auth0/auth0-react/blob/master/EXAMPLES.md#2-protecting-a-route-in-a-gatsby-app

### Is there a way to disable this screen when logging in for the first time?   

We can live with it for the time being.

Screen Shot 2021-08-21 at 9 33 50 AM

src/pages/edit.js
vnugent: we can improve it in another PR

src/components/auth0NavBar/Auth0Profile.js
vnugent: we can move this component to the top nav bar

vnugent (approved):

PR #147, opened 2021-12-31
Area page additions

gibboj:
* adding in basic info to area page
* Moving icons away from direct import next didn't like it, I didn't want to deal with another package
* SEO > SeoTags
* turning off lint checks


gibboj:
Basic pages for areas & climbs. I separated `areas with subareas` from `areas with climbs` and added the latter as "crag". This removes all the `if (hasChildren) / else` situation in leaf-area-page

vnugent:
thanks for bringing the app back to life!

PR #148, opened 2022-01-01
Bugfix: score for grade

gibboj:
Now includes V grades & doesn't break on bad matches


vnugent (approved):

vnugent:
Do we also want these changes in `nextjs` branch as well?  I'm going to merge to `develop` for now

Issue #146, opened 2021-12-31
Fix unit test in Next branch


vnugent:
- [x] fix unit tests
- [x] fix linting errors

PR #149, opened 2022-01-01
Clean up deps, fix jest and linting

vnugent:


vnugent:
Fixing https://github.com/OpenBeta/open-tacos/issues/146

Issue #160, opened 2022-01-04
Missing climb type (trad, sport, bouldering, etc)


vnugent:
sample page: https://open-tacos.vercel.app/climbs/c89d1225-e8ee-4592-829c-e814d7b803b3

<img width="790" alt="Screen Shot 2022-01-04 at 12 42 44 PM" src="https://user-images.githubusercontent.com/3805254/148108195-043285fc-d4d8-467f-95cb-4a20998c5cb4.png">

vnugent:
Same issue here https://github.com/OpenBeta/open-tacos/blob/nextjs/src/pages/crag/%5Bid%5D.tsx#L95

---
 
<img width="930" alt="Screen Shot 2022-01-10 at 6 07 22 PM" src="https://user-images.githubusercontent.com/3805254/148868892-91cbb032-9d6b-45e3-9ed2-8d750177809d.png">

PR #183, opened 2022-01-15
Missing climb type in card and climb page

vnugent:
Fixing  https://github.com/OpenBeta/open-tacos/issues/160


vnugent:
Screen Shot 2022-01-15 at 11 25 13 AM
---
Screen Shot 2022-01-15 at 11 25 27 AM

PR #210, opened 2022-02-07
Cluster fix

gibboj:


src/pages/index.tsx
```diff
@@ -94,7 +94,7 @@ export const getStaticProps: GetStaticProps = async ({ params }) => {
           comparison: 'gt'
         }, {
           field: 'density',
-          num: 0.005,
+          num: 0.5,
```
gibboj, line 90: Just my opinion but we were getting a lot of states on the explore page which I think is too broad to be of interest.

vnugent: fine with me

vnugent:
I know you have this in progress.  For #236  I had to make a lot of changes to the backend's aggregate object (to return more data) and it breaks the Area page: grade distribution histograms and map.  I'm going to comment them out and fix them in another PR.

Issue #214, opened 2022-02-14
Missing crag/wall name in climb page breadcrumbs


vnugent:
### Bug
Breadcrumbs should display the wall/crag name 

`Home / US / Utah / Southeast Utah / Indian Creek / Supercrack Buttress`

[This line](https://github.com/OpenBeta/open-tacos/blob/develop/src/components/ui/BreadCrumbs.tsx#L25) is probably responsible.

<img width="495" alt="Screen Shot 2022-02-14 at 2 53 33 PM" src="https://user-images.githubusercontent.com/3805254/153877351-3e3c30ce-d0cd-4326-a4ca-f860614fe7f8.png">

PR #226, opened 2022-03-01
Add mini breadcrumps to climb card

vnugent:
also fix issue #214


vnugent:
Screen Shot 2022-03-01 at 9 10 55 PM

Issue #178, opened 2022-01-13
M2 release: Crag finder


vnugent:
## Story

As a sport / trad climber, I want to find crags or a destination with a high number of climbs that match my ability.

Important filters: 
* [x] High density of climbs for my ability
* [ ] Temps/seasons
* [ ] Sport | Trad filter
* [ ] Star ratings ?

Todo:
* Design
  - [x] Dev prototype #174 
  - [ ] Designer layout
* Frontend
  - [x] New search component: TBD
* Backend: 
   - [x] https://github.com/OpenBeta/openbeta-graphql/issues/47

vnugent:
Screenshots of work in progress

### Autocomplete search by city/landmark/point of interest

show crag counts vs driving time below each suggestion

<img width="600" alt="Screen Shot 2022-03-15 at 7 47 58 AM" src="https://user-images.githubusercontent.com/3805254/158323482-33040673-b4bf-4fcf-ac7e-ddee13e390ea.png">

---

### Result results + filter bar

<img width="750" alt="Screen Shot 2022-03-15 at 7 48 26 AM" src="https://user-images.githubusercontent.com/3805254/158323898-9dab84d9-3ec2-43fc-92de-6f443f6140a3.png">

vnugent:
### YDS Filter
<img width="883" alt="Screen Shot 2022-03-24 at 8 37 31 PM" src="https://user-images.githubusercontent.com/3805254/159996766-ae1a28c9-33e7-4649-b807-abc2ab10ca2b.png">

### Distance filter

<img width="714" alt="Screen Shot 2022-03-24 at 5 44 25 PM" src="https://user-images.githubusercontent.com/3805254/159996812-e176b3be-5dc7-4cfc-aebb-6d4a98cdd8f1.png">

vnugent:
Released!

PR #231, opened 2022-03-17
Add map to search results page

vnugent:
Task #178 
Show search results on map.  Mose interactions to be done in future PRs.
- [x] upgrade map libs
- [x] simplify Heatmap component


vnugent:
Screen Shot 2022-03-17 at 10 12 54 AM

PR #233, opened 2022-03-20
fix: #230 correcting the breadcrumb link for the crag url when on the climb page

### LINKED ISSUE #230: Breadcrumbs link to wrong URL
Crag link is incorrectly point to `areas/<id>`  (it should be `crag/<id>`) 

Code:  https://github.com/OpenBeta/open-tacos/blob/develop/src/components/ui/BreadCrumbs.tsx#L33
Suggestion:  if it's the last element link to `crag/id`

---

Screen Shot 2022-03-09 at 12 42 52 PM

clintonlunn:
This fixes #230 where the breadcrumb links to /areas/{id} instead of /crag/{id} while on the climb page

i basically added another condition where if both `isLastElement` and `isClimbPage` are both true, then use the `/crag/`in the url, and the rest of the previous breadcrumb urls will continue to use `/areas/`. 

with fix you can see the url being hovered over is now going to link to `/crag/{id}` (my pointer is not getting captured in the screenshot, but you can see the url being hovered over in bottom left of screenshot and the hovered breadcrumb is underlined):

image

all previous breadcrumb link beyond the last link to /areas/{id}:
image

Please let me know if there are improvements that can be made. Thanks!

### DIFF
--- a/src/components/ui/BreadCrumbs.tsx
+++ b/src/components/ui/BreadCrumbs.tsx
@@ -35,27 +35,26 @@ function BreadCrumbs ({ pathTokens, ancestors, isClimbPage = false }: BreakCrumb
         return (
           <span key={index}>
             <span className='text-secondary mx-1.5'>/</span>
-            {isLastElement && !isClimbPage
-              ? (
-                <span className=''>{sanitizeName(place)}</span>
-                )
-              : isLastElement && isClimbPage
-                ? (
-                  <span className='text-secondary'>
-                    <Link href={climbPageLastUrl}>
-                      <a className='hover:underline hover:text-gray-900'>
-                        {sanitizeName(place)}
-                      </a>
-                    </Link>
-                  </span>
-                  )
-                : (
-                  <span className='text-secondary'>
-                    <Link href={url}>
-                      <a className='hover:underline hover:text-gray-900'>{sanitizeName(place)}</a>
-                    </Link>
-                  </span>
-                  )}
+            {/* last element but not currently on the climb page */}
+            {isLastElement && !isClimbPage && <span className=''>{sanitizeName(place)}</span>}
+
+            {/* last element and currently on the climb page, so /crag/ url used */}
+            {isLastElement && isClimbPage &&
+              <span className='text-secondary'>
+                <Link href={climbPageLastUrl}>
+                  <a className='hover:underline hover:text-gray-900'>
+                    {sanitizeName(place)}
+                  </a>
+                </Link>
+              </span>}
+
+            {/* all links before last element when on climb page to get /areas/ url */}
+            {!isLastElement && isClimbPage &&
+              <span className='text-secondary'>
+                <Link href={url}>
+                  <a className='hover:underline hover:text-gray-900'>{sanitizeName(place)}</a>
+                </Link>
+              </span>}
           </span>
         )
       })}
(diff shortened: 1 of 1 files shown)


src/components/ui/BreadCrumbs.tsx
```diff
@@ -31,20 +31,31 @@ function BreadCrumbs ({ pathTokens, ancestors, isClimbPage = false }: BreakCrumb
...
+                    </Link>
+                  </span>
+                  )
+                : (
+                  <span className='text-secondary'>
+                    <Link href={url}>
+                      <a className='hover:underline hover:text-gray-900'>{sanitizeName(place)}</a>
+                    </Link>
+                  </span>
+                  )}
```
MichaelDimmitt: nested ternary is usually a bad idea

consider replacing all of the ternaries when logic becomes complex to use short circuiting statements
```JavaScript
{ isLastElement && isClimbPage && <componentForTrue> }
{ !(isLastElement && isClimbPage) && <componentForFalse> }

```

vnugent:
@allcontributors bot please add @clintonlunn  @MichaelDimmitt for code and ideas

clintonlunn:
@vnugent Hi there, is there anything else you would like to see or any changed made in this PR before it gets merged?

vnugent:
As I pointed out in the [previous comment](https://github.com/OpenBeta/open-tacos/pull/233/files/a4e4215f437c187480322a84ccf7a13b971f697a#diff-c6c8479f405b7d6a500cdddbea7ea6ba4f4654b92fbe4295b2345c4c7eecb529R52), the check on line 52 causes blank breadcrumbs when we're viewing areas/[id].tsx and crag/[id].tsx

you can see the breakage in the CI build:
- https://open-tacos-7efd0drf5-openbeta-dev.vercel.app/areas/62276601f794104b6821600d
- https://open-tacos-7efd0drf5-openbeta-dev.vercel.app/crag/62276601f794104b68216053

These checks were there so that we don't wrap `a href=' '` around the last entry when we're already on that page.  I think the complexity is probably not worth it for little UX gain.  If you can just simplify that section by removing all the extra checks, I'll merge it. 
Something like this:

```
     return (
          <span key={index} className='text-secondary'>
            <span className='mx-1.5'>/</span>
            <Link href={isLastElement && isClimbPage ? climbPageLastUrl : url}>
                 <a className='hover:underline hover:text-gray-900'>
                    {sanitizeName(place)}
                  </a>
            </Link>
          </span>
   )
```

clintonlunn:
@vnugent I agree, it does remove a lot of complexity if everything just gets an href. The change you suggested seems to work just fine on my local machine. ~~The Vercel docs say that each push triggers a new build to happen, but I don't see it/am looking in the wrong place. Also just not familiar with Vercel in the first place.~~ I see Vercel has now deployed a new instance of the app.

clintonlunn:
Actually, it looks like there is still an issue with the breadcrumbs. 

If you are 1 level up from the climb page, i.e. a subarea, the link still points to /areas/{id} (where before it was not a link, but a span element. Even though it adds some complexity, maybe this is a good argument for accounting for it? 
See for example: 
image

vnugent:
I'd be happy to accept whichever solution you come up that works for all 3 cases (climb page, crag page and all ancestor area pages).
ps: I've also relaxed a setting in Vercel to allow CI builds from fork repos.  You should probably receive a notification from Vercel after a git push.

clintonlunn:
Great, I will think on it and see if I can come up with a more concise solution that accounts for all 3 scenarios. Thanks for taking the time to review this!

clintonlunn:
@vnugent not as readable as I'd like it to be, but I believe it covers all the cases we've talked about and a little more concise than my first attempt. For clarification, when you say 'all ancestor area pages', do you mean the parent pages of the crag pages?

vnugent:
> , when you say 'all ancestor area pages', do you mean the parent pages of the crag pages?
yes, I meant all the non-crag/area pages accessible.  It's good to go.  Thanks for the PR!

PR #245, opened 2022-03-28
Climbing discipline popover (#243)

### LINKED ISSUE #238: Add Bouldering grade filter popover
### Crag Finder story #178 
Use case:  as a boulderer I want to find bouldering areas with problems within my grade.

#### Tips:
The V grade component should be very similar to YDS filter and Radius filter. 
- YDS: https://github.com/OpenBeta/open-tacos/blob/develop/src/components/YDSFilter.tsx
- Radius: https://github.com/OpenBeta/open-tacos/blob/develop/src/components/RadiusFilter.tsx

Screen Shot 2022-03-24 at 8 37 31 PM

### LINKED ISSUE #243: Convert button group into popover
Move buttons into a popover.  See YDSFilter.tsx and RadiusFilter.tsx

Screen Shot 2022-03-26 at 8 58 40 PM

clintonlunn:
fixes #243 any feedback appreciated! 

- make a portion of `<FilterPopover />` props optional to accommodate the `<DisciplineFilter />`
- move `<DisciplineGroup />` into it's own component file
- move discipline filtering into a popover
- add add click handler for Cancel button
image

### DIFF
--- a/src/components/DisciplineGroup.tsx
+++ b/src/components/DisciplineGroup.tsx
@@ -1,26 +1,35 @@
+import { useEffect, useState } from 'react'
 import { ClimbDisciplineRecord } from '../js/types'
 import { FilterToggleButton } from './ui/Button'
 
 interface DisciplineGroupProps {
+  defaultTypes: Partial<ClimbDisciplineRecord>
   climbTypes: Partial<ClimbDisciplineRecord>
   onChange: Function
 }
 
-const DisciplineGroup = ({ climbTypes, onChange }: DisciplineGroupProps): JSX.Element => {
+const DisciplineGroup = ({ climbTypes, onChange, defaultTypes }: DisciplineGroupProps): JSX.Element => {
+  const [types, setTypes] = useState(defaultTypes)
+
+  useEffect(() => {
+    if (JSON.stringify(climbTypes) !== JSON.stringify(defaultTypes)) {
+      onChange({ ...climbTypes, sport: types.sport, tr: types.tr, trad: types.trad, bouldering: types.bouldering })
+    }
+  }, [])
   return (
     <div className='flex space-x-2'>
       <FilterToggleButton
         selected={climbTypes.sport}
         label='Sport'
         // eslint-disable-next-line
-        onClick={() => onChange({ ...climbTypes, sport: !climbTypes.sport })}
+        onClick={() => onChange({ ...types, sport: !types.sport })}
       />
       <FilterToggleButton
         selected={climbTypes.trad}
         label='Trad'
         onClick={() => {
           // eslint-disable-next-line
-          onChange({ ...climbTypes, trad: !climbTypes.trad })
+          onChange({ ...climbTypes, trad: !types.trad })
         }}
       />
       <FilterToggleButton
@@ -28,7 +37,7 @@ const DisciplineGroup = ({ climbTypes, onChange }: DisciplineGroupProps): JSX.El
         label='Top rope'
         onClick={() => {
           // eslint-disable-next-line
-          onChange({ ...climbTypes, tr: !climbTypes.tr })
+          onChange({ ...climbTypes, tr: !types.tr })
         }}
       />
       <FilterToggleButton
@@ -36,7 +45,7 @@ const DisciplineGroup = ({ climbTypes, onChange }: DisciplineGroupProps): JSX.El
         label='Bouldering'
         onClick={() => {
           // eslint-disable-next-line
-          onChange({ ...climbTypes, boulder: !climbTypes.bouldering })
+          onChange({ ...climbTypes, bouldering: !types.bouldering })
         }}
       />
     </div>
--- a/src/components/DisciplineGroup.tsx
+++ b/src/components/DisciplineGroup.tsx
@@ -1,36 +1,52 @@
-import { actions, cragFiltersStore } from '../js/stores'
+import { useState } from 'react'
+import { cragFiltersStore } from '../js/stores'
 import { FilterToggleButton } from './ui/Button'
 
-const DisciplineGroup = (): JSX.Element => {
+const DisciplineGroup = ({ onChange }): JSX.Element => {
   const { trad, sport, tr, boulder } = cragFiltersStore.useStore()
+  const [sportSelected, setSportSelected] = useState(sport)
+  const [tradSelected, setTradSelected] = useState(trad)
+  const [trSelected, setTrSelected] = useState(tr)
+  const [boulderSelected, setBoulderSelected] = useState(boulder)
+
+  console.log({ sportSelected, tradSelected, trSelected, boulderSelected })
   return (
     <div className='flex space-x-2'>
       <FilterToggleButton
-        selected={sport}
+        selected={sportSelected}
         label='Sport' onClick={() => {
-          void actions.filters.toggle('sport')
+          setSportSelected(!sportSelected)
+          onChange('sport')
         }}
       />
       <FilterToggleButton
-        selected={trad}
+        selected={tradSelected}
         label='Trad' onClick={() => {
-          void actions.filters.toggle('trad')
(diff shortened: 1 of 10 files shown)


vnugent:
I've just checked out the CI build. Looking great!  Thanks.

Can you make 1 change to keep toggle states **local** in the DisciplineFlter?  This allows users to cancel the changes if they want (clicking on Cancel or elsewhere outside the popover), and only write changes to the data store when users click 'Apply'

Here's how YDS filter maintains local min & max
https://github.com/OpenBeta/open-tacos/blob/develop/src/components/YDSFilter.tsx#L11-L16

clintonlunn:
@vnugent Sounds good, I will make that change!

src/components/ui/Button.js
```diff
@@ -21,7 +21,7 @@ export function IconButton ({ onClick, children, active, text, className }) {
 
 export const FilterToggleButton = ({ selected, label, onClick }) => {
   return (
-    <button type='button' onClick={onClick} className={`border-2 rounded-2xl  btn-small active:scale-95 ${selected ? 'border-neutral-100 text-neutral-100' : 'border-neutral-500 text-neutral-400 hover:border-neutral-400'}`}>
+    <button type='button' onClick={onClick} className={`border-2 rounded-2xl  btn-small active:scale-95 ${selected ? 'border-neutral-800 text-neutral-800' : 'border-neutral-400 text-neutral-400 hover:border-neutral-600'}`}>
       {label}
     </button>
   )
```
MichaelDimmitt, line 27: why are the other files tsx? but Button is .js?

src/components/ui/FilterPopover.tsx
```diff
@@ -25,7 +25,7 @@ const FilterPopover = ({ label, header, children, onApply, min, max }: FilterPop
       >
         <header className='mb-16'>{header}</header>
         <div className='px-4'>{children}</div>
-        {min !== undefined && max !== undefined &&
+        {(min !== undefined && max !== undefined) && min !== undefined && max !== undefined &&
```
MichaelDimmitt: is this duplicated logic?

src/components/ui/Button.js
vnugent: we migrated to TS 2 months ago and not everything was ported

clintonlunn:
@vnugent i believe i was able to get the behavior you asked for. any feedback appreciated

clintonlunn:
> I've just checked out the CI build. Looking great! Thanks.
> 
> Can you make 1 change to keep toggle states **local** in the DisciplineFlter? This allows users to cancel the changes if they want (clicking on Cancel or elsewhere outside the popover), and only write changes to the data store when users click 'Apply'
> 
> Here's how YDS filter maintains local min & max https://github.com/OpenBeta/open-tacos/blob/develop/src/components/YDSFilter.tsx#L11-L16

@vnugent i believe i was able to get the behavior you asked for. any feedback appreciated

src/components/DisciplineGroup.tsx
```diff
@@ -0,0 +1,46 @@
+import { useState } from 'react'
+import { cragFiltersStore } from '../js/stores'
+import { FilterToggleButton } from './ui/Button'
+
+const DisciplineGroup = ({ onChange }): JSX.Element => {
+  const { trad, sport, tr, boulder } = cragFiltersStore.useStore()
```
MichaelDimmitt: does it need to be dynamic?
```javascript
const items = { trad, sport, tr, boulder } // these are boolean values.

const [pillDetails, setPillDetails] = useStore(Object.values(items));

return (
  <div className='flex space-x-2'>
    Object.keys(items).map( (pill, index) => 
        <FilterToggleButton
            key={index}
            selected={pillDetails[index]}
            label={pill} onClick={() => {
            let tempPillDetails = pillDetails
            tempPillDetails[index] = pillDetails[index]
            setPillDetails([...tempPillDetails])
            onChange(pill)
        }}
      />
   ) 
)
```
sorry if this is formatted bad

wrote this up in the really small github window, probably should have messed with the browser styles or did in an ide.

Screen Shot 2022-04-01 at 11 47 43 AM

vnugent:
I'm glad to see it's working. Many thanks! This is a summary of what I commented inline.  If possible can you please do some additional refactoring:

### DisciplineFilter.tsx
- Instead of using an array of string flags for `climbTypes`, I'd use an object of type [ClimbDisciplineRecord](https://github.com/OpenBeta/open-tacos/blob/develop/src/js/types.ts#L31).  This way we can just toggle a state directly, eg `climbTypes.sport = true | false` without the need for a loop/util function.
- No need to undo changes with `onCancel()` since changes are local to this  component.  The Cancel button's default behavior is to close the popup.   As long as we don't update the global filter settings by calling actions.filters.\<update Disciplines\>, the global settings won't change.

### DisciplineGroups.tsx
- You can maintain local states here, but don't access the global store directly for the initial states.  Instead, pass the initial states in via a prop, eg `defaultTypes` (type=ClimbDisciplineRecord)
```javascript
const DisciplineGroup = ({ defaultTypes, onChange }): JSX.Element => {
   const [types, setTypes] = useState(defaultTypes)
   return (
      <FilterToggleButton
          selected={types.sport}
           label='Sport' 
           onClick={() => {
              onChange({...types, sport: !types.sport})
           }}
      />
    ...
```

MichaelDimmitt:
nice,
you could map over Object.keys(types) to generate <FilterToggleButton/>
though if this component will never have more pills or never have less pills it may not be needed.

vnugent:
> nice, you could map over Object.keys(types) to generate though if this component will never have more pills or never have less pills it may not be needed.

I was about to respond to your review comment above.  Yes, we can certainly do that.  There are 7 types

```javascript
export type ClimbDiscipline = 'sport' | 'bouldering' | 'alpine' | 'tr' | 'trad' | 'mixed' | 'aid'
```

what do you think if we improve it in a future PR?   

ps: watch out for "bouldering" vs "boulder" [issue](https://github.com/OpenBeta/open-tacos/issues/250)

clintonlunn:
@MichaelDimmitt @vnugent Thanks to you both for your criticism and detailed suggestions. I was able to put most of the suggestions into practice with the exception of the local state management within `<DisciplineGroup />`. Instead, I used the climbTypes state that is in the `<DisciplineFilter />` component to dictate which pill button is active/inactive.

After lots of troubleshooting, I came up with this solution use of a useEffect that was able to reset the active/inactive to be in sync with the defaultValues. Without it, the values would not correctly reset after cancelling the filters. 

However, I recognize that there is probably a better solution than this. Any comments and suggestions definitely welcome.

vnugent:
@clintonlunn Thanks so much for the updates.  Good workaround!

It wasn't as simple as I thought.  When the popover is canceled/hidden, its parent filter component is not dismounted and therefore still maintains previous values.  I just got lucky with the YDS and Radius filters thanks to the 3rd-party slider component.

If headlessui [`Popover`](https://headlessui.dev/react/popover) had an onClose(), we could use it to reset those values.

src/js/stores/index.ts
```diff
@@ -229,10 +229,13 @@ export const cragFiltersStore = createStore('filters')({
...
-      sport === true ? set.sport(true) : set.sport(false)
-      trad === true ? set.trad(true) : set.trad(false)
-      tr === true ? set.tr(true) : set.tr(false)
-      bouldering === true ? set.boulder(true) : set.boulder(false)
+      // Update multiple states at once
+      api.set.state(draft => {
+        draft.sport = sport === true
+        draft.trad = trad === true
+        draft.tr = tr === true
+        draft.boulder = bouldering === true
```
MichaelDimmitt, line 237: draft.boulder = Boolean(bouldering)

MichaelDimmitt: or if it is already a boolean
and we know it will always be one 
```
draft.boulder = bouldering
```

PR #254, opened 2022-04-03
add tests to DisciplineGroup and simplify cancel useEffect #243

clintonlunn:
- simplify how cancelling works
- add some tests to verify that the DisciplineGroup popover works as expected

### DIFF
--- a/src/components/DisciplineGroup.tsx
+++ b/src/components/DisciplineGroup.tsx
@@ -11,7 +11,7 @@ interface DisciplineGroupProps {
 const DisciplineGroup = ({ climbTypes, setClimbTypes, defaultTypes }: DisciplineGroupProps): JSX.Element => {
   useEffect(() => { // if doesn't match, set everything to default
     if (JSON.stringify(climbTypes) !== JSON.stringify(defaultTypes)) {
-      setClimbTypes({ ...climbTypes, sport: defaultTypes.sport, tr: defaultTypes.tr, trad: defaultTypes.trad, bouldering: defaultTypes.bouldering })
+      setClimbTypes({ ...defaultTypes })
     }
   }, [])
 
--- a/src/components/__tests__/DisciplineGroup.tsx
+++ b/src/components/__tests__/DisciplineGroup.tsx
@@ -1,0 +1,59 @@
+import React from 'react'
+import '@testing-library/jest-dom/extend-expect'
+import { act, render } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+
+import DisciplineGroup from '../DisciplineGroup'
+
+const defaultTypes = {
+  trad: true,
+  sport: true,
+  tr: true,
+  bouldering: true
+}
+
+let climbTypes = {
+  trad: true,
+  sport: true,
+  tr: true,
+  bouldering: true
+}
+
+const setClimbTypes = (value): void => {
+  climbTypes = value
+}
+
+describe('DisciplineGroup', () => {
+  it('renders component without error', () => {
+    render(<DisciplineGroup defaultTypes={defaultTypes} climbTypes={climbTypes} setClimbTypes={setClimbTypes} />)
+  })
+
+  it('all 4 types selected by default', async () => {
+    const { container, getByRole, rerender } = render(<DisciplineGroup defaultTypes={defaultTypes} climbTypes={climbTypes} setClimbTypes={setClimbTypes} />)
+    expect(container.getElementsByClassName('border-neutral-800').length).toBe(4)
+
+    const sportButton = getByRole('button', { name: /Sport/i })
+    console.log(sportButton)
+
+    await userEvent.click(sportButton)
+
+    await act(async () => {
+      await new Promise((resolve) => setTimeout(resolve, 3))
+    })
+    rerender(<DisciplineGroup defaultTypes={defaultTypes} climbTypes={climbTypes} setClimbTypes={setClimbTypes} />)
+    expect(container.getElementsByClassName('border-neutral-800').length).toBe(3)
+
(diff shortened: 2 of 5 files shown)


MichaelDimmitt (approved):

src/components/DisciplineGroup.tsx
```diff
@@ -11,7 +11,7 @@ interface DisciplineGroupProps {
 const DisciplineGroup = ({ climbTypes, setClimbTypes, defaultTypes }: DisciplineGroupProps): JSX.Element => {
   useEffect(() => { // if doesn't match, set everything to default
     if (JSON.stringify(climbTypes) !== JSON.stringify(defaultTypes)) {
-      setClimbTypes({ ...climbTypes, sport: defaultTypes.sport, tr: defaultTypes.tr, trad: defaultTypes.trad, bouldering: defaultTypes.bouldering })
+      setClimbTypes({ ...defaultTypes })
```
clintonlunn, line 14: I believe this might be good simplification for resetting the climbTypes

vnugent:
Glad to see unit tests!  To make those toggle buttons more testable and A11y-friendly, consider adding `aria-checked` attr and change `role` to `checkbox` in [FilterToggleButton](https://github.com/OpenBeta/open-tacos/blob/develop/src/components/ui/Button.js#L22):
```javascript
<button type='button' role='checkbox' aria-checked={selected}  ... />
```

then we can assert its state
```javascript
const sportButton = getByRole('button', { name: /Sport/i })

expect(sportButton).toBeChecked()  
// or expect(sportButton).not.toBeChecked()  


```

src/components/__tests__/DisciplineGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const { rerender: rerender2, getByRole: getByRole2 } = render(<DisciplineGroup defaultTypes={defaultTypes} climbTypes={climbTypes} setClimbTypes={setClimbTypes} />)
+
+    // rerender needed so that useEffect [] runs.
+    rerender2(<DisciplineGroup defaultTypes={defaultTypes} climbTypes={climbTypes} setClimbTypes={setClimbTypes} />)
+
+    const sportButton2 = getByRole2('checkbox', { name: /Sport/i })
+    const tradButton2 = getByRole2('checkbox', { name: /Trad/i })
+    const topRopeButton2 = getByRole2('checkbox', { name: /Top rope/i })
+    const boulderingButton2 = getByRole2('checkbox', { name: /Bouldering/i })
```
MichaelDimmitt, line 85: you should not need to redeclare these right?
we re-rendered but it is still the same element I would imagine it updated?

src/components/__tests__/DisciplineGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+      await userEvent.click(button)
+      rerender(<DisciplineGroup defaultTypes={defaultTypes} climbTypes={climbTypes} setClimbTypes={setClimbTypes} />)
+    }
+
+    const { container, getByRole, rerender, unmount } = render(<DisciplineGroup defaultTypes={defaultTypes} climbTypes={climbTypes} setClimbTypes={setClimbTypes} />)
+    expect(container.getElementsByClassName('border-neutral-800').length).toBe(4)
+
+    const sportButton = getByRole('checkbox', { name: /Sport/i })
+    const tradButton = getByRole('checkbox', { name: /Trad/i })
+    const topRopeButton = getByRole('checkbox', { name: /Top rope/i })
```
MichaelDimmitt, line 54: can you still get by role button but expect .toBeChecked?
that was what the comment on the pr mentioned.

did you find out get by checkbox to be best practices when researching?

src/components/__tests__/DisciplineGroup.tsx
clintonlunn: it seems that the original climbType buttons still have the old values even after explicitly unmounting the component, which is why i redeclared the variables. from what i can tell these are just stale values. do you think that could indicate an issue?

src/components/__tests__/DisciplineGroup.tsx
clintonlunn: cannot `getByRole('button')` because button is not a type here, checkbox is. the ask was to change role to checkbox. i figured it was a small typo in declaring the new sportButton variable in the comment above. 
```javascript
const sportButton = getByRole('button', { name: /Sport/i })
// should have been
const sportButton = getByRole('checkbox', { name: /Sport/i })
```

regarding best practices, i didn't find anything that argued against using this as a way to get the button.  [getByTestId](https://testing-library.com/docs/queries/bytestid/) might also be good

MichaelDimmitt: get by test ID is frowned upon.

src/components/__tests__/DisciplineGroup.tsx
MichaelDimmitt: as long as you checked, fine by me

MichaelDimmitt: you can change the const to let then and reassign those buttons

that would make it more copy and paste friendly.
however, having seperate consts is safer.

there are merrits to both ways.

these are just suggestions to get you to think about it. feel free to disregard.

src/components/__tests__/DisciplineGroup.tsx
clintonlunn: I see that now after reading through the docs some more 🤦‍♂️:

"In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use {getByTestId} only after the other queries don't work for your use case."

PR #259, opened 2022-04-09
Add boulder grade filter popover #238

### LINKED ISSUE #238: Add Bouldering grade filter popover

clintonlunn:
- [x] add bouldering filter range ui components
- [x] add BOULDER_DEFS with correct properties/values
- [x] move filter components into filter folder
- [x] add cragFilterStore boulderRange state value and mutation methods 
- [x] working map filtering
- [x] add mobile view props

### DIFF
--- a/src/js/grades/rangeDefs.ts
+++ b/src/js/grades/rangeDefs.ts
@@ -70,3 +70,63 @@ export const YDS_DEFS: SliderMarksType =
     score: getScoreForSort('5.15', GradeScales.Yds)
   }
 }
+
+export const BOULDER_DEFS: SliderMarksType =
+{
+  0: {
+    score: 0,
+    label: 'V-easy'
+  },
+  1: {
+    score: 5,
+    label: 'V1'
+  },
+  2: {
+    label: 'V2',
+    score: getScoreForSort('5.0', GradeScales.VScale)
+  },
+  3: {
+    label: 'V3',
+    score: getScoreForSort('5.5', GradeScales.VScale)
+  },
+  4: {
+    label: 'V4',
+    score: getScoreForSort('5.6', GradeScales.VScale)
+  },
+  5: {
+    label: 'V5',
+    score: getScoreForSort('5.7', GradeScales.VScale)
+  },
+  6: {
+    label: 'V6',
+    score: getScoreForSort('5.8', GradeScales.VScale)
+  },
+  7: {
+    label: 'V7',
+    score: getScoreForSort('5.9', GradeScales.VScale)
+  },
+  8: {
+    label: 'V8',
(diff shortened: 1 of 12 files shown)


clintonlunn:
@vnugent I think I'm somewhat close here. Could you take a look and see if anything jumps out at you? I cannot get the same number of crags as I was getting before implementing this bouldering grade filter popover. It seems that it was displaying all bouldering areas. Now that there's a grade restriction on it, there are less crags (-8 less in the area I've been testing in, which is California). Maybe this is some bad data being omitted now?

Also, do we want to have separate methods for updating free vs bouldering climbing? i.e. like the methods i added in [index.ts](https://github.com/OpenBeta/open-tacos/pull/259/files#diff-d6e49eb2aa96fcb8102b97710b982cb62152810794869ba2d935d43469facebd)

vnugent:
if you haven't done it already, install [Apollo Client Devtool](https://www.apollographql.com/docs/react/development-testing/developer-tooling/#apollo-client-devtools) browser extension, which allows you to browse cache data coming from the backend.  
FWIW, the graphql api returns all data by radius. No sport/trad/bouldering filtering is done there.

src/js/grades/rangeDefs.ts
```diff
@@ -70,3 +70,67 @@ export const YDS_DEFS: SliderMarksType =
...
+  },
+  12: {
+    label: 'V11',
+    score: getScoreForSort('V11', GradeScales.VScale)
+  },
+  13: {
+    label: 'V12',
+    score: getScoreForSort('v12', GradeScales.VScale)
+  },
+  14: {
```
clintonlunn, line 132: This is a part I am unsure about. I don't think will capture everything above v13+, but rather up to V13+. I am not sure how to account for this? Perhaps increase number of tick marks to account for v grades up to what is currently the max. A quick google suggests the highest grade out there is a v17 (https://www.99boulders.com/bouldering-grades)

clintonlunn:
There is still a question about whether the filtering is happening correctly. I understand that the graphql requests happen only to select by radius and the other filtering methods only act on what has been retrieved in the radius based search.

vnugent:
We can do more testing and address the data issue in a future PR. I'll have a look at the PR later.  Thanks!

Issue #253, opened 2022-04-03
Remove bookmark heart symbol until functionality works


clintonlunn:
As has been previously discussed, remove the bookmark heart symbol until functionality works

vnugent:
Completed in PR https://github.com/OpenBeta/open-tacos/pull/260

Issue #249, opened 2022-03-30
Mobile navigation 


vnugent:
Create a dedicate mobile layout to simplify the current navbar

PR #260, opened 2022-04-12
Show map on mobile

vnugent:
- [x] Improve map view on mobile/tablet
- [x] Partially fix #269 
- [x] https://github.com/OpenBeta/open-tacos/issues/249
- [x] https://github.com/OpenBeta/open-tacos/issues/253
- [x] Simplify global layout classes


vnugent:
Introduce Map view and List view

Screen Shot 2022-04-13 at 6 16 39 PM

---

Screen Shot 2022-04-13 at 6 17 01 PM

clintonlunn:
this looks really nice on smaller screens! however, on a larger screen (25'' 2560 x 1440 (16:9)) , there is a lot of whitespace. See below:

image
image

image

Could we consider expanding the navbar (also possibly footer?) and the home hero to full width and showing 4 columns in the "Explore" section with a slightly larger container width for  these larger screens? I'm not sure if there's already a design drawn up for this, don't want to step on any toes!

vnugent:
Good point. We can also consider a two-column design with curated content such as Explore, Recommendations on the left, and a heat map on the right, similar to the crag finder result page which is now full screen.

vnugent:
I wanted to add tests to verify mobile vs desktop layout but unable to get jest mocking to work. 

https://github.com/OpenBeta/open-tacos/issues/271

PR #267, opened 2022-04-15
stack route stat components in  RouteCard (#203)

### LINKED ISSUE #203: Handle climb card overflow better
**Actual**
Screen Shot 2022-02-04 at 15 42 55

**Expected**
Climbing discipline label shouldn't overlap with grade label

clintonlunn:
There is more room for both of these components (RouteGradeGip, RouteTypeChips) if they are stacked on top of eachother. 
image

This also seems to handle larger climb type children here too. See below for rendering all chips (just for testing).
image

mobile view:
image

### DIFF
--- a/src/components/ui/Chip.tsx
+++ b/src/components/ui/Chip.tsx
@@ -21,7 +21,7 @@ function Chip ({ type }: ChipProps): JSX.Element {
   return (
     <span
       aria-label='climb-discipline'
-      className={`font-extralight font-mono rounded-sm py-1 mr-4 px-2 text-xs uppercase border-2 ${ChipType[type]}`}
+      className={`font-extralight font-mono rounded-sm py-1 mr-3 px-2 my-1 text-xs uppercase border-2 ${ChipType[type]}`}
     >
       {type}
     </span>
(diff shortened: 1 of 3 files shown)


clintonlunn:
Hi @vnugent is there anything holding this pr up or any changes you would like me to make??

vnugent:
Can you fix the typo 'RouteGradeGip'?

vnugent:
I commented inline earlier.  Not sure if you got the notification.

clintonlunn:
@vnugent apologies, I did not see your comment. I cannot find it either, can you link to your original inline comment?

Also, I fixed that typo.

vnugent:
I can't link to it.  Here's a screencap

---
Screen Shot 2022-04-19 at 9 37 28 AM

clintonlunn:
@vnugent, no I never saw this, nor do I see it on the pull request at all even now. I think you need to complete the review and request changes if needed for me to be able to see your comments.

vnugent:
Ah my bad.  My review was still in 'Pending'.  I think that explains why you didn't see it.

PR #274, opened 2022-04-20
replace randomly generated 'climbs for you' value with actual total

### LINKED ISSUE #248: Calculate number of climbs for you
Story #178 
### Use case:
As a new climbs I can only climb up 5.8.  I want to quickly identify crags with a high number of "climbs for me".

The piechart is currently displaying a random value.  We need to calculate the count based on selected disciplines and grade range. The logic is the same as how highlighted cells are determined in https://github.com/OpenBeta/open-tacos/blob/develop/src/components/ui/DTable.tsx#L44

One way to go about this is to create a "selector" in the data store similar to [this](https://github.com/OpenBeta/open-tacos/blob/develop/src/components/ui/DTable.tsx#L44).

Having a selector also allows us to sort matching crags by number of of climbs for me.

---
### Screenshot
Screen Shot 2022-03-29 at 8 50 12 PM

clintonlunn:
addressing #274. there are other issues that are visible in the following screenshots, but i created separate issues (see #276 and #275) to address. 

previously, there was a randomly generated number populating the 'climbs for you' section of each crag row. I replaced it with an actual total that is calculated from the difficulty bands.

image
image

### DIFF
--- a/src/components/finder/CragRow.tsx
+++ b/src/components/finder/CragRow.tsx
@@ -5,9 +5,11 @@ import { AreaType } from '../../js/types'
 import DTable from '../ui/DTable'
 import { MiniCrumbs } from '../ui/BreadCrumbs'
 import useResponsive from '../../js/hooks/useResponsive'
+import { cragFiltersStore } from '../../js/stores'
 
 /* eslint-disable-next-line */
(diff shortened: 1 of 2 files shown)


vnugent:
Thanks for the PR.  The pie chart is now useful.

clintonlunn:
> Thanks for the PR. The pie chart is now useful.

@vnugent not quite, i only corrected the number inside of the pie chart. that work will be completed in #276

PR #282, opened 2022-04-22
Difficulty tables #275

### LINKED ISSUE #275: the difficulty table does not display boulder difficulty correctly
the difficulty table does not display boulder difficulty correctly, instead it displays the difficulty range for trad, sport, or tr. in the below screenshots you can see that the boulder difficulty table doesn't match.

where the difficulty range gets set: https://github.com/OpenBeta/open-tacos/blob/develop/src/components/ui/DTable.tsx#L21

image

image

clintonlunn:
The difficulty tables now represent their respective disciplines. 

image

### DIFF
--- a/src/components/ui/__tests__/DTable.tsx
+++ b/src/components/ui/__tests__/DTable.tsx
@@ -2,27 +2,28 @@ import React from 'react'
 import '@testing-library/jest-dom/extend-expect'
 import { render, screen } from '@testing-library/react'
 
-const byDisciplineAggBoulder = {
+const byDisciplineAggBoulderSport = {
   sport:
     {
       total: 10,
       bands: {
-        beginner: 6,
+        beginner: 4,
         intermediate: 6,
         expert: 0,
         advance: 0
       }
     },
   trad: null,
-  boulder: {
-    total: 3,
-    bands: {
-      beginner: 2,
-      intermediate: 0,
-      expert: 0,
-      advance: 1
-    }
-  },
+  boulder:
+    {
+      total: 3,
+      bands: {
+        beginner: 2,
+        intermediate: 0,
+        expert: 0,
+        advance: 1
+      }
+    },
   tr: null
 }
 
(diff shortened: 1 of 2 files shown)


clintonlunn:
I am attempting to create some tests for the DTable component, but running into errors I cannot seem to resolve.

image

vnugent:
@clintonlunn Since DTable depends on the global data store, we need to use Jest 'mock' to return a 'fake' data store, bypassing the actual GraphQL API call.

vnugent:
I just pushed some changes to make mocking work. Using mocks as per the [official documentation](https://jestjs.io/docs/es6-class-mocks) won't work, because we use ts-jest without babel.  Some helpful links:

- Jest mock without babel https://github.com/facebook/jest/issues/10025#issuecomment-716789840
- How to select a single row in react testing: https://testing-playground.com/gist/782aae09bf284b7c6c04b09129753878/e1b93ac1a2274a2898e372bc5e9b39ccb101ba60

clintonlunn:
@vnugent thanks so much for taking the time to add this. I really appreciate it! I will get some additional tests added for this and also this will be a good point of reference to test other components in this project!

src/components/ui/__tests__/DTable.tsx
```diff
@@ -0,0 +1,82 @@
...
+import '@testing-library/jest-dom/extend-expect'
+import { render, screen } from '@testing-library/react'
+
+const byDisciplineAggBoulder = {
+  sport:
+    {
+      total: 10,
+      bands: {
+        beginner: 6,
+        intermediate: 6,
```
vnugent, line 11: @clintonlunn can you change `intermediate` to  6 to make the total 10?

vnugent (approved):

src/components/ui/__tests__/DTable.tsx
vnugent: I mean '4' (4 + 6 = 10)

clintonlunn:
@vnugent I made the correction you asked for. Also added an additional test to make sure that the difficulty table re-highlights. might be a little bit of a trivial test, but a good learning exercise to see how you set these tests up! if you think there are any additional cases to cover, please let me know!

PR #284, opened 2022-04-23
Bring back heatmap layer to give users more context

vnugent:
(Re)introduce the heatmap layer to give users more context of what there is in the area regardless of user's filters.  The heatmap layer can also be used for other pages.

Screen Shot 2022-04-23 at 3 03 16 PM

---

### close up view (zoomed in)
- turn off heatmap
- show areas not in user preferences in smaller font and marker

Screen Shot 2022-04-23 at 3 05 35 PM

### DIFF
--- a/src/components/maps/MarkerLayer.tsx
+++ b/src/components/maps/MarkerLayer.tsx
@@ -1,12 +1,11 @@
 import { Properties, FeatureCollection } from '@turf/helpers'
 import { Source, Layer, LayerProps } from 'react-map-gl'
 
-export const LayerId = 'area-labels'
-
 const layerStyle: LayerProps = {
-  id: LayerId,
+  id: 'my-areas',
   type: 'symbol',
   source: 'areas',
+  filter: ['==', 'isInMyRange', true],
   layout: {
     'icon-size': ['interpolate', ['linear'], ['zoom'], 8, 1.25, 12, 1],
     'symbol-spacing': 5,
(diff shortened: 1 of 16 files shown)


src/components/maps/MarkerLayer.tsx
```diff
@@ -1,28 +1,59 @@
...
+    'text-size': ['interpolate', ['linear'], ['zoom'], 8, 12, 14, 10],
+    'icon-allow-overlap': false
+  },
+  paint: {
+    'icon-color': '#ffffff',
+    'text-halo-blur': 4,
+    'text-halo-width': 2,
+    'text-color': '#a3a3a3',
     'text-halo-color': '#334455'
+    // 'text-opacity':
```
MichaelDimmitt: css variables could be used
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties

to centralize all of your colors.

MichaelDimmitt:
Is there a way to test that it is operating correctly?

Though maybe we do not need to test the inner workings of 3rd party libraries.
I could go either way.

src/components/maps/MarkerLayer.tsx
vnugent: Thanks.  I opened https://github.com/OpenBeta/open-tacos/issues/287

vnugent:
@MichaelDimmitt we should be at least testing the global store, but we're not at the moment.  For custom map components I guess we can mock Mapbox/webgl coomponents and use jest snapshots for the rest?

Issue #216, opened 2022-02-14
Add official colors from style guide


vnugent:
### TODO
Define official colors (style guide link TBD) in tailwind config:

https://github.com/OpenBeta/open-tacos/blob/develop/tailwind.config.js#L18-L22

Ali7040:
I hope this message finds you well. I would like to kindly request the assignment of the following issue to myself. Your consideration in assigning this issue to me would be greatly appreciated.

vennelaeruvala:
I would like to contribute to this issue. Please consider allowing me to do it

sun-flame:
i would like to contribute. please consider me to do it.

sarthakjena05:
Hi, I would like to contribute. Please consider me to do it.

shashank-poola:
Hey, I would like to contribute.

vnugent:
This task is poorly defined.  New task TBD.

Akshay-Chougala:
Hi, I'm new to open source and would love to work on this issue. Can I take this up?

Vk2006-blip:
Please assign this task to me. I’ll define the official colors in the Tailwind config and update them once the style guide link is available

priyapulakhandam:
Hey @vnugent i really want to try and contribute towards this

meet-innovate:
Hi! I’m new to the project and would love to work on this issue if it’s still open.
Please let me know if I can get started. Thanks!

PR #285, opened 2022-04-23
subtract forYou values from total to make pie chart make sense

### LINKED ISSUE #276: "Climbs For You" pie chart does not display correct information
the pie chart in the "Climbs For You" section of a crag finder page does not seem to display correct information. 

image

clintonlunn:
addresses issue #276 

gray represents the leftover climbs that are not in the desired range, and the green highlighted are the climbs that are within the desired range. As per @vnugent's comment down below, I also changed the color of the pie chart to match the difficulty table colors.

**Before**:
image

**After**:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "husky": "^7.0.4",
     "jest": "^27.5.1",
     "postcss": "^8.4.5",
+    "react-test-renderer": "^17.0.0",
     "tailwindcss": "^3.0.23",
     "ts-jest": "^27.1.4",
     "ts-standard": "^11.0.0",
--- a/src/components/ui/Statistics/CounterPie.tsx
+++ b/src/components/ui/Statistics/CounterPie.tsx
@@ -21,7 +21,7 @@ const CounterPie = ({ total, forYou }: CounterPieProps): JSX.Element => {
           outerRadius='80%'
           startAngle={90}
           endAngle={450}
-          fill='#62cae3'
+          fill='#8BB191B3'
           dataKey='value'
           paddingAngle={5}
           isAnimationActive={false}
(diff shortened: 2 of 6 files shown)


MichaelDimmitt:
hmm, what was the before behavior?

clintonlunn:
> hmm, what was the before behavior?

@MichaelDimmitt good point, I should have included the before behavior! It basically displayed (total / forYou), so in the first row it would be (3 / 2), which would not make too much sense to the user. 

Before behavior:
image

MichaelDimmitt:
I see you did not fix the number,
you fixed the pie chart to show an appropriate ratio,

you get claps 👏

vnugent:
@clintonlunn do you think we should change the color of the active pie to light green, same as user band color on the distribution table?

clintonlunn:
@vnugent Ok, I changed the colors to match the difficulty table (see above in original post for screenshot). I also matched the opacity values from the difficulty tables (to 70% opacity). What do you think?

Also, I added a snapshot test to verify that the piecharts were correctly rendered, which meant I had to add an additional dev dependency. I hope that's okay! Let me know if you don't think it serves much purpose and I will remove it.

vnugent:
> Also, I added a snapshot test to verify that the piecharts were correctly rendered, which meant I had to add an additional dev dependency. I hope that's okay! Let me know if you don't think it serves much purpose and I will remove it.

Fine with me.  And thanks for writing tests.  I think snapshot testing have their place when when we can't easily test D3 or WebGL components.

vnugent:
I've just realized the low contrast between the gray the secondary color makes it a bit hard to see when the number is small.  But let's revisit colors/theming later in another PR  :)

---
Screen Shot 2022-04-24 at 8 27 49 PM

clintonlunn:
> I've just realized the low contrast between the gray the secondary color makes it a bit hard to see when the number is small. But let's revisit colors/theming later in another PR :)
> 
> Screen Shot 2022-04-24 at 8 27 49 PM

@vnugent agreed the contrast is pretty low. There is a current issue open in #216 that it can probably be addressed in? To me this would be probably also not be compliant from a a11y standpoint.

PR #293, opened 2022-04-27
Use next/img to improve Lighthouse score

vnugent:


vnugent:
- Image size:  next/img isn't helpful here because it's still receiving raw, full-size jpg image URLs from wikimedia ad flickr.  When we move to our own image service, the API will return properly sized images.

PR #299, opened 2022-05-05
Add auth0 and login/logout buttons

### LINKED ISSUE #298: Setup authentication
We need to add site-wide authentication for 2 important upcoming features:
- Photo uploading #176
- Adding/editing content #177 

### Tasks:
- [ ] Integrate https://github.com/nextauthjs/next-auth
- [ ] Pick an Auth provider (Auth0, Okta, etc)
- [ ] Create a simple login page

We want to be able to allow authenticated users to call backend APIs so something like OAuth + JWT could work.

nthh:
Fixes #298 

Adds in auth0, and a Login/logout button on the mobile more popover and at the top navbar on larger devices. 

Uses [Auth0's Next.js SDK](https://github.com/auth0/nextjs-auth0/) and their Universal Login page (redirects to their site for login) but we could change this to use a custom login page. 

On localhost there's a dialog shown below, but enabling "Allow Skipping User Consent" should skip this for prod. 
 
Screen Shot 2022-05-04 at 8 03 22 PM

## Todo 
Add this to env variables (one app for dev, one for prod)
```
AUTH0_SECRET=use [openssl rand -hex 32] to generate a 32 bytes value
AUTH0_BASE_URL=http://localhost:3000
AUTH0_ISSUER_BASE_URL=https://auth0domain
AUTH0_CLIENT_ID=CLIENT_ID
AUTH0_CLIENT_SECRET=CLIENT_SECRET
```

Configure dev/prod apps (one tenant for prod, one for dev) in auth0
Go to Applications -> APIs and enable "Allow Skipping User Consent" ([more info](https://auth0.com/docs/get-started/applications/confidential-and-public-applications/user-consent-and-third-party-applications#revoke-consent))
Go to Authentication -> Passwordless and enable email passwordless.
Go to Connections under both applications settings pages and disable everything except for passwordless
Add allowed callback/logout urls for each application (e.g. http://localhost:3000/api/auth/login for callback and http://localhost:3000 for logout)
Go to Branding -> Universal Login -> Login and set the template to be passwordless and add ` passwordlessMethod: 'link'` into the template:
  ```
},
      passwordlessMethod: 'link',
      configurationBaseUrl: config.clientConfigurationBaseUrl,
 ````
Set up branding for login page/emails (logo, any changes to signin/registration email templates)
For prod, add CNAME that maps auth.tacos.openbeta.io to AUTH0_DOMAIN.auth0.com.


vnugent:
> Configure dev/prod apps (one tenant for prod, one for dev) in auth0

@nthh is this something @nadr0 or I have to do it in Auth0 correct?

> For prod, add CNAME that maps auth.tacos.openbeta.io to AUTH0_DOMAIN.auth0.com.

I'll take care of this today

vnugent:
Error page when I clicked on the CI [Preview](https://open-tacos-git-nthhsetup-auth-298-openbeta-dev.vercel.app/)

Would Nextjs [Auth solution](https://github.com/nextauthjs/next-auth) make it easier to integrate with our project?

Screen Shot 2022-05-05 at 1 56 33 PM

nthh:
> @nthh is this something @nadr0 or I have to do it in Auth0 correct?

Yes, or I can do it once I have access. 


>  For prod, add CNAME that maps auth.tacos.openbeta.io to AUTH0_DOMAIN.auth0.com.

This isn't necessary, but just gives users extra confidence on the redirect that they are still on a domain managed by OpenBeta.




> Would Nextjs [Auth solution](https://github.com/nextauthjs/next-auth) make it easier to integrate with our project?

You're seeing an error because the environment variables aren't configured in that environment, and next-auth will be similar implementation to this. We could use next-auth but I thought it made sense to use auth0's SDK. 


A few things to note:
As the client secret is required since auth is done server-side ([Authorization Code Flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow)), it is not wise to include the client secret in the public repo. Therefore, we should either create a new Auth0 app in the dev tenant for each contributor who needs access to auth or give them the secret for the dev app. 

Alternatively, we could use a client-only library, but that is only recommended in the following scenarios: 
> You are using [Static HTML Export](https://nextjs.org/docs/advanced-features/static-html-export) with Next.js.
> You do not need to access user data during server-side rendering.
> You want to get the access token and call external API's directly from the frontend layer rather than using Next.js API Routes as a proxy to call external APIs

Also, Auth0 does not allow you to edit the email templates (with custom branding, etc) unless you use an external email provider.

vnugent:
Do you think it makes sense to use [client-side](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce) auth?

- The bulk of our pages are public-facing.  We use Static Generation (SSG) at build time to pre-generate those pages, not SSR on user request.  And maybe with the exception of the user account page(s) which can be client-side only, do you foresee the need for SSR pages with private user data?
- If we go with client-side, I'm not sure how to protect photo upload API calls to our 3rd-party photo management provider which uses simple auth.

nthh:
> Do you think it makes sense to use [client-side](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce) auth?

> do you foresee the need for SSR pages with private user data?

We should be able to use client-side auth, but I don't have much experience with SSR/Next.js or an intuition about what could go wrong if we don't go with server-side auth. 

> * If we go with client-side, I'm not sure how to protect photo upload API calls to our 3rd-party photo management provider which uses simple auth.

The easiest way I see to do this is by having a server-side function (lambda or similar) check a JWT provided by the client (and signed by our auth provider) and then do whatever is needed by the storage provider (and return a signed asset URL) if the JWT is valid.

vnugent (changes requested):
What do you think if we make Login the primary call-to-action (solid button) and Discord secondary (same color, but outlined button)?

---

Screen Shot 2022-05-07 at 9 16 19 PM

nthh:
> 

That's fine with me. What about logout button? And I'll implement client-side auth this week if you think we should move forward with that

vnugent:
Let's go ahead with the client-side auth.  Does it make sense if we turn the "Login" link into a drop down menu once the user has signed in?  The first, default option would be their public facing profile screen (path=/u/\<user id\>), the last option Logout

nthh:
Pushed latest changes. I'm not impressed by auth0's React library, as the way it manages cookies/reauthorizes seems slow (note the time it takes for the navbar to show up on refresh after authenticated) and I don't trust the callbacks to actually be parsed and not require a refresh to take effect (I've seen this but can't reproduce it). In addition, it won't really work at all on Safari or Brave browser unless we use custom domains (which makes us upgrade to the paid plan on auth0). But it should get the job done for now and makes configuration much easier as opposed to managing secrets and such.

nthh:
I'll add the vercel dev app to auth0 for now so it can be tested without running locally

vnugent:
I've just tested the preview build.
- After clicking on the link in the email, my email client (Thunderbird) launches OpenTacos in a new tab in Brave.  The Login button is still active.  I'll have to click on it again to see the Logout button.  Is it the normal behavior? 
*update:

>  In addition, it won't really work at all on Safari or Brave browser unless we use custom domains (which makes us upgrade to the paid plan on auth0).

Ah you did mention Brave.  I know it's painful now to keep changing, but maybe it's worth going with nextjs auth + Auth0, or even switch to a completely different provider? eg Okta.

nthh:
> Ah you did mention Brave. I know it's painful now to keep changing, but maybe it's worth going with nextjs auth + Auth0, or even switch to a completely different provider? eg Okta. 

At this point I would rather use nextauth with a db we manage. I think we'll have the same issues with a different provider such as Okta and third-party cookies on certain browsers. 

Unless you have any ideas about how to make nextjs auth0 work with managing secrets in a public repo. Here are my potential solutions ranking from easy/less secure to harder/more secure:

1. Make a dev tenant and prod tenant. Expose the dev tenant's secret in the repo
2. Make a dev tenant and prod tenant. Provide the secret to new developers in private
3. Make a dev tenant and prod tenant. Make a new application in the dev tenant for each contributor

vnugent:
Can we use Okta, Auth0, or the like as the provider and [next-auth.js](https://next-auth.js.org/) lib + **PKCE**?

Please correct me if I'm wrong.  From what I read, PKCE is suitable for SPA or mobile where we can't securely store the client secret.  In the Okta snippet (link below), I don't see the client secret being used. (I assume the config would be similar for Auth0)

https://github.com/nextauthjs/next-auth/issues/1530#issuecomment-800563809

If 3rd-party cookies + PKCE is an issue, we can look into paying for a custom domain.  I think the extra costs outweigh hosting our own user DB at this stage for reasons we've previously discussed in chat, ie, maintenance and security.

nthh:
Thanks for that link, it gave me the confidence to push deeper and see what our options were and I figured out a solution. A PKCE security check is [already used](https://github.com/nextauthjs/next-auth/blob/db188b872f8efb90fc325dc60644941dc0650e92/packages/next-auth/src/providers/auth0.ts#L19) by next-auth (along with [state](https://auth0.com/docs/secure/attack-protection/state-parameters)) as a security measure to verify the client and secure the request against certain attacks. 

From what I've gathered, this means that the client secret is not necessary and our **token_endpoint_auth_method** can be set to **none**. I can't find many instances of this being done with any of the node.js libraries as I guess our use case is unique but this should work. See the description in the openid client that next-auth uses [here](https://github.com/panva/node-openid-client/blob/v4.6.0/docs/README.md#client-authentication-methods).

PR #309, opened 2022-05-12
fixed NextJS compilation warning re stylesheet

CocoisBuggy:
no ref needed in layout component. _app.tsx already provides it as its parent

Notable is [This doc](https://nextjs.org/docs/basic-features/built-in-css-support#adding-a-global-stylesheet) which say imports at this level are global


vnugent:
@allcontributors bot please add @CocoisBuggy as a contributor for code, ideas

Issue #289, opened 2022-04-24
Sporadic error causing blank page (Maximum update depth exceeded)


vnugent:
I encountered this error now and then.  I think it originated from the CragsMap component when I was viewing a highly populated area such as Boulder or Red Rock, and at the same time trying to zoom in and out when the component was still rendering.

https://reactjs.org/docs/error-decoder.html/?invariant=185

---

<img width="1127" alt="Screen Shot 2022-04-24 at 8 20 52 PM" src="https://user-images.githubusercontent.com/3805254/164991676-724c3d4d-c5e3-45b2-8f56-d856bd8dd60b.png">

clintonlunn:
I am seeing this issue happen anytime that the user pans around the CragsMap

vnugent:
A temporary fix could be wrapping the CragsMap component with an error boundary component https://www.npmjs.com/package/react-error-boundary

CocoisBuggy:
After #311 I can't replicate this issue anymore

vnugent:
Thanks.  I'm closing the issue.

PR #311, opened 2022-05-13
Fixed infinite recurse when panning map

CocoisBuggy:
https://visgl.github.io/react-map-gl/docs/get-started/state-management#controlled-map

Before, you could not pan the map at all. Now the behavior seems to be as expected.

(Also added type annotation to getMapDivDimensions)

### DIFF
--- a/src/components/layout.tsx
+++ b/src/components/layout.tsx
@@ -16,10 +16,7 @@ function Layout ({ contentContainerClass = 'content-fullscreen-tablet', rootCont
         <title>Climbing Route Catalog</title>
         <meta name='description' content='Open license climbing route catalog' />
         <link rel='icon' href='/favicon.ico' />
-        <link
-          href='/fonts/fonts.css'
-          rel='stylesheet'
-        />
+
         <SeoTags
           keywords={['openbeta', 'rock climbing', 'climbing api']}
           description='Climbing route catalog'
--- a/src/components/maps/BaseMap.tsx
+++ b/src/components/maps/BaseMap.tsx
@@ -22,10 +22,6 @@ interface BaseMapProps {
     width: number
     height: number
   }
-  initialViewState: ViewState & {
-    width: number
-    height: number
-  }
   onViewStateChange: (event: ViewStateChangeEvent) => void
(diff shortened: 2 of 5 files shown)


vnugent:
The map is not maximized full height when I navigate away then go back.

To reproduce:
1. Search for Las Vegas
2. Click on an area dot on the map to bring up the popup
3. Click on the popup
4. Browser go back.  The map div is now shorter (screenshot)

---

Screen Shot 2022-05-13 at 1 57 46 PM

vnugent:
Are you seeing this problem?  https://github.com/OpenBeta/open-tacos/issues/289

CocoisBuggy:
#289 is the issue that this was supposed to fix, let me look into the responsive container

CocoisBuggy:
Oh, that is not the correct solution. That reverts back to the infinite recurse

vnugent:
> https://github.com/OpenBeta/open-tacos/issues/289 is the issue that this was supposed to fix, let me look into the responsive container
FYI, the MapBox GL component requires a pre-defined height hence the need for the hook to calculate the height.

CocoisBuggy:
Bit of a head-scratcher

They have an internal event for listening to resizes [here](https://github.com/visgl/react-map-gl/blob/bb9f1f45d845421fef8c39c3b6c39a3426db3e52/src/mapbox/mapbox.ts#L565-L568) although, of course, this is just a wrapper for the mapbox-gl library... so it is really difficult to nail down exactly why it skips over our modification of the height prop.

```tsx
const mapRef = useRef(null)

  // Map seems to struggle responding to height changes after initial set
  useEffect(() => {
    if (mapRef.current !== null) {
      const map: MapboxMap = (mapRef.current as any)
      map.resize()
    }
  }, [height, mapRef])

  return (
    <Map
    ref={mapRef}
    ...
    />
```

There is a potential problem here; being that when resize events are dispatched resize() will be called twice. I would still expect it to be about as performant as it was before...

vnugent:
Thanks for looking into this.  I've verified the preview build on large screen and mobile devtool.

PR #312, opened 2022-05-15
Coco/custom404

CocoisBuggy:
Nice simple 404 page that fulfils the base spec.

At the very least, it's not a blank page with invisible error message (The current solution in place)

Bit of fuss with git reversion, looks like 404.js did not get properly migrated. If that history is important, I can go back and do the steps in the correct order.


src/pages/404.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import Header from '../components/Header'
+
+function NotFoundPage (): JSX.Element {
+  return (
+    <>
+      <div className='absolute w-full h-screen bg-gray-50 overflow-hidden p-2'>
```
vnugent: I think `h-screen` pushes the page footer out of the viewport.

src/pages/404.tsx
```diff
@@ -0,0 +1,34 @@
...
+              <div className='bg-yellow-100 italic mb-2'>
+                404
+              </div>
+            </div>
+
+            <div className='bg-white text-4xl md:text-7xl tracking-wide font-bold px-2 py-1'>
+              Page not found
+            </div>
+
+            <div className='mt-2 bg-white'>
```
vnugent: can we leave out the background (bg-white) to let the div inherit whatever is set higher up?  

Screen Shot 2022-05-15 at 11 11 55 AM

vnugent (changes requested):

src/pages/404.tsx
CocoisBuggy: Footer was left out, I have changed from manual usage of header to <Layout />

src/pages/404.tsx
CocoisBuggy: Sure thing

PR #314, opened 2022-05-16
Coco/layout changes

CocoisBuggy:
Looking for feedback on these proposals.

Some stuff is different from the first one (In the area explorer)


vnugent:
Looking great so far!  The thing that catches my attention the most is the auto-zoom action on mouse click.  Even if I'm familiar with an area, the closeup view takes the overall context away from me, I'd have to zoom out again to see where it is in relation to other areas.   I'd suggest only highlight/emphasize the map label/marker.

Take a look at this AirBnb map for Barcelona.  The map remains static which helps me orient each property in relation to each other and well-known landmarks.

https://www.airbnb.com/s/Barcelona--Spain/homes?place_id=ChIJ5TCOcRaYpBIRCmZHTz37sEQ

Just my opinion.  I hope others will chime in.

CocoisBuggy:
I think that makes a lot of sense. I'm not super familiar with U.S Geography so I suppose I'm equally disoriented zoomed in or out.
Let me get rid of the auto-zoom (and also the need for two clicks per nav-action). I think it'll make the navigation faster anyway.

It also occurs to me that a map animation may be more jarring than smooth on devices with more limited hardware

CocoisBuggy:
# Notes from this weeks meeting
**Just putting these here so that they don't get lost**

###  Area View
1. Find a way to disclose more information to the user up front without forcing them to click down the hierarchy to get more information about child areas. (I will try and find a spin on this that retains the ability to show more detailed information about the current area without deviating too harshly from the ideas laid out in the finder page)
2. Ideally, we want a singular map component that is flexible enough to be used in both places (Or at least cuts down on too much duplication)
3. The map is supposed to load areas dynamically. This will need an efficient API polyfill to keep redundant calls down for network throttled devices, and may require an algo to group densely distributed markers into just the label of their parent areas. I will lay out two possible approaches to this end with one being a client-side implementation utilizing data stores, and the other being a server side algo which may ultimately end up being more performant depending on how many nodes map-gl can reasonably accommodate.

###  Crag View
1. We  want to provide the user with the option to rapidly return to a map-centric view of the data. My approach to this will be to load neighbour crags ad-hoc when the user asserts the change UNLESS the user has come to the crag view from its parent area view - in which case that initial context data can be passed horizontally.
2. Same story with disclosure being too gradual, will experiment with some potential solutions.
3. Images need to be addressed.

### Route (Climb) view
_See figma spec._

### Summary
Try to not have too many divergent control schemes for the website. The user should be able to intuit the behavior of one explorer based on another one -views may be visually different to accommodate different entity types, so long as the top-down UI patterns are the same.

## Upshot
I am going to abandon this P.R for the time being and approach this problem from the chore refactoring that needs to happen in order to generalize our core explorer components.

_If I've forgotten anything, please add it below this comment_

vnugent:
Thanks for the summary. It's very helpful.  I'll give a proper response later.  
May I suggest that we create a rough wireframe prototype in Figma?

CocoisBuggy:
That sounds very sensible to me

vnugent:
I've put together a mockup for the climb page view on larger screen.  As I said before, while we're lacking some useful content such as photos and star ratings, we can make up for it by playing our strength, which is giving users better maps.

*edit:  click on the Chevron button to minimize/maximize the panel.

https://www.figma.com/proto/Lvtb9bcbbN8zBexVSvaaFh/OpenBeta-Prototypes?node-id=225%3A379&scaling=scale-down&page-id=0%3A1&starting-point-node-id=225%3A379&show-proto-sidebar=1

---

Area view (default)

Screen Shot 2022-05-25 at 4 26 06 PM

Crag row expanded

Screen Shot 2022-05-25 at 4 26 47 PM

CocoisBuggy:
I like this concept, I'm sure I'll have some ideas while implementing so I'll report back if I think of anything that might be done differently

PR #332, opened 2022-06-01
Wide changes addressing unfinished pages

CocoisBuggy:
Header props have been extended.
Climb, Area, Crag explorers have been touched up with some new ideas
Media allowances have been made on area / crag / climb. Will still need further testing.

### DIFF
--- a/src/components/area/areaMap.tsx
+++ b/src/components/area/areaMap.tsx
@@ -41,8 +41,7 @@ function getBounds (areas: Mappable[]): [[number, number], [number, number]] | n
       [minLon, minLat], // SouthWest corner
       [maxLon, maxLat]] // northeastern corner of the bounds
   }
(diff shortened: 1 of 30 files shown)


src/components/area/areaMap.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+interface AreaMapProps {
+  subAreas: Mappable[]
+  area: AreaType
+  focused: string | null
+  selected: string | null
+}
+
+/** get the bounds needed to display this map */
+function getBounds (areas: Mappable[]): [[number, number], [number, number]] | null {
```
vnugent, line 23: I think we have a utility that calculates bounding box for a list of geojson.  We should try to re-use it in future PR.

vnugent (approved):

vnugent:
Thanks for this huge PR!

Issue #318, opened 2022-05-19
Better UX for next-auth pages


nthh:
Next-auth pages are unbranded and confusing. For example, if a user clicks on an email link that has already been used [this page](https://tacos.openbeta.io/api/auth/signin?error=Callback) is shown. Any similar pages should be evaluated and improved using [custom pages](https://next-auth.js.org/configuration/pages).

PR #366, opened 2022-06-15
Custom sign in page for Next Auth

jstruzik:
Partially handles https://github.com/OpenBeta/open-tacos/issues/318.

I'm not sure the best place to get style review, but here's my shot at it. I'm an awful designer so please be gentle 😄 

image


vnugent:
I think it's way better than the default page

PR #369, opened 2022-06-17
Adding left/right navigation for keyboards and phones

jstruzik:
I thought it would be useful to be able to easily navigate through a crag from left to right when navigating a climb page. I made the assumption that gestures (keyboard and swiping) would be the best way of doing this but perhaps we use visual elements instead/in addition (left and right arrows).

Some notes:

- Uses two new react hook libraries for the gesture detection
- ~~I'm wondering if there's a way to speed up/cache these pages more aggressively. Navigation should feel pretty seamless~~ (NextJS seems to do a good job of this in production)
- ~~I'm running into warning from GQL about cache key conflicts for the `ClimbMeta` type. I couldn't quite understand how to make the cache keys unique~~ (resolved in https://github.com/OpenBeta/open-tacos/pull/369/commits/797ed3e0b8cc342f64c6b7cd1747e266cacd798a) (or even better, cache the entire area's sorted climbs higher up for other pages to use).


jstruzik:
Example page: https://open-tacos-git-left-right-climb-nav-openbeta-dev.vercel.app/climbs/046bae28-b231-55cf-8c00-69a57cef75f4

vnugent:
Great idea!  Thanks.

For the cache issue, you can see how it's being done for area metadata https://github.com/OpenBeta/open-tacos/blob/develop/src/js/graphql/Client.ts

Maybe you can add a `ClimbMeta` entry and climbId as the key to see if those warnings go away?  if not we can address it in another PR.

jstruzik:
> Maybe you can add a ClimbMeta entry and climbId as the key to see if those warnings go away?

That seemed to have worked, thanks!

vnugent:
> (or even better, cache the entire area's sorted climbs higher up for other pages to use).

Glad you figured out how to get sibling climbs. Would having next/previous sibling fields in the get-climb-by-ID query be useful?  we can definitely add it to the backend to save us from making the additional query.  Though I'd be happy to accept the PR as it is.

vnugent:
@jstruzik Can you please merge/rebase with `develop` to pick up the margin/layout fix PR? then we're good to go.

Issue #415, opened 2022-07-08
Allow users to provide an optional website link


vnugent:
### Use case
As a climber/photographer sharing photos on opentacos, I want to provide a way for other users to contact me (website, Instagram link, etc)

### Todo
- [ ] Email viet at openbeta.io for Auth0 keys 
- [ ] Add a new optional textfield to profile edit form https://github.com/OpenBeta/open-tacos/blob/develop/src/components/users/account/ProfileEditForm.tsx
- [ ] Add a validation rules (max length and valid characters)

### Mock up

<img width="691" alt="Screen Shot 2022-07-08 at 9 35 40 AM" src="https://user-images.githubusercontent.com/3805254/177947578-855a420f-dcaa-44e3-abd1-5c0a9280dd8c.png">

PR #421, opened 2022-07-10
Allow users to provide an optional website link #415

xavier-kong:


vnugent:
can you also add the website link to the public-facing component (PublicProfile.tsx) and wrap it with the html `a` tag.

https://github.com/OpenBeta/open-tacos/blob/develop/src/components/users/PublicProfile.tsx#L25

src/js/utils.ts
```diff
@@ -165,3 +165,12 @@ export const saveAsFile = (data: any, filename: string): void => {
...
   saveAs(blob, filename)
 }
+
+/**
+ * Website URL validation
+ * @param url
+ * @returns true if valid website URL
+ */
+export const checkWebsiteUrl = (url: string) => {
+  return !url.includes(' ') && url.length > 2
```
vnugent: Can you add a stricter validation using RegEx?  Thanks! 

Something like this StackOverflow suggestion:  https://stackoverflow.com/a/68780191

vnugent (changes requested):

PR #448, opened 2022-07-28
Functional Favourite system with scoped collections.

CocoisBuggy:
This is maybe not my best code of all time, but this will solve the immediate problem.

State hoisting in some places might make it  more performant, but not by a significant margin

### DIFF
--- a/src/components/area/panel/panelList.tsx
+++ b/src/components/area/panel/panelList.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 import ListItem, { ListItemEntity } from './listItem'
 
 interface PanelListProps {
@@ -12,6 +12,16 @@ interface PanelListProps {
 }
 
 export default function PanelList (props: PanelListProps): JSX.Element {
+  const [favs, setFavs] = useState<string[]>([])
+  useEffect(() => {
+    fetch('/api/user/fav')
+      .then(async res => await res.json())
+      .then(collections => {
+        setFavs(collections.areaCollections.favourites)
+      })
+      .catch(console.error)
+  }, [])
+
   useEffect(() => {
     if (props.selected !== null) {
       const selectedElement = document.getElementById(`${props.selected}-xqoops98`)
@@ -49,6 +59,7 @@ export default function PanelList (props: PanelListProps): JSX.Element {
           key={item.id}
         >
           <ListItem
+            isFav={favs.includes(item.id)}
             onFocus={reFocusCheck}
             onClick={reClickCheck}
             selected={props.selected === item.id}
--- a/src/components/area/panel/panelList.tsx
+++ b/src/components/area/panel/panelList.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import { useEffect, useState } from 'react'
+import { APIFavouriteCollections } from '../../../pages/api/user/fav'
 import ListItem, { ListItemEntity } from './listItem'
 
 interface PanelListProps {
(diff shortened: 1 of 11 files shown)


vnugent:
@CocoisBuggy can you see why the build is failing?  Probably `collections` isn't initialized or something.

src/components/users/FavouriteButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface Props {
+  climbId?: string
+  areaId?: string
+}
+
+function NoLogin (): JSX.Element {
+  return (
+    <button
+      disabled
+      className='text-center p-2 px-3 border-2 rounded-xl border-rose-500 transition
```
vnugent: may I suggest that we use a [brand color](https://github.com/OpenBeta/open-tacos/blob/2a5e5aa82f282f89892375625117a7920414577e/tailwind.config.js#L20-L23)?

vnugent (changes requested):

src/components/users/FavouriteButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+  // If there is some kind of programming error / user is un-authenticated we render the default
+  // interaction-devoid button
+  if ((climbId === undefined && areaId === null) || session.status === 'unauthenticated') {
+    return <NoLogin />
+  }
+
+  return (
+    <button
+      disabled={loading}
+      onClick={toggle}
```
vnugent, line 96: Right now when the user is not logged in, the button is a no-op.  What if we remove the NoLogin button?  when the user isn't logged in and click on it, the onClick() calls NextAuth `LogIn()` function, similar to the main login button?

vnugent (changes requested):

src/components/users/FavouriteButton.tsx
CocoisBuggy: Seems sensible, will change

src/components/users/FavouriteButton.tsx
CocoisBuggy: Makes sense, have  changed to primary color

vnugent:
Climbs are correctly added.  

Screen Shot 2022-07-31 at 12 49 12 PM

But when I un-favorite a climb and then refresh the browser.  The change is not written to the backend.  There are still 3 climbs and the climb is shown as favorite as before.

CocoisBuggy:
Sorted, just a silly mistype over here https://github.com/OpenBeta/open-tacos/blob/80a4d591768d512f8c0117b820b7ae8e192d7afb/src/pages/api/user/fav.ts#L84

The adding / removing should be symmetrical now

src/components/area/panel/panelList.tsx
```diff
@@ -12,6 +12,16 @@ interface PanelListProps {
 }
 
 export default function PanelList (props: PanelListProps): JSX.Element {
+  const [favs, setFavs] = useState<string[]>([])
+  useEffect(() => {
+    fetch('/api/user/fav')
+      .then(async res => await res.json())
+      .then(collections => {
+        setFavs(collections.areaCollections.favourites)
```
vnugent: I think `favs` is undefined here (for users who have added areas to their favorite) causing a blank screen.

Screen Shot 2022-07-31 at 4 16 00 PM

To reproduce:

1.  Log in Auth0 and remove user's metadata `collections` (existing users won't have this entry)
2. Visit an area page such as https://open-tacos-dbdv8vk67-openbeta-dev.vercel.app/areas/47667307-c0a0-5f50-873d-2a5c6d757fb2

vnugent (changes requested):

PR #449, opened 2022-07-29
WIP remove image via Sirv api

amandril:


vnugent:
@all-contributors add @amandril for code and ideas

vnugent:
I've opened a support ticket with Sirv re API permission.

src/components/media/slideshow/SlideViewer.tsx
```diff
@@ -202,22 +203,27 @@ const InfoContainer = ({ currentImage, tagList, auth }: InfoContainerProps): Rea
...
 
       {tagList?.length === 0 && isAuthorized &&
         <div className='my-8 text-secondary flex items-center space-x-1'>
           <LightBulbIcon className='w-6 h-6 stroke-1 stroke-ob-primary' />
           <span className='mt-1 text-xs'>Your tags help others learn more about the crag</span>
         </div>}
+
+      {isAuthorized &&
+        <div className='my-8'>
+          <RemoveImage imageInfo={currentImage} tagCount={tagList?.length} />
```
vnugent: Perhaps introduce a callback eg 'onImageDeleted()` in RemoveImage component.  The function is called after a successful delete.  You can handle the callback here in SliderViewer and call onClose() which will bubble the event up to the gallery component that will take care of closing the full screen viewer.

src/components/media/RemoveImage.tsx
```diff
@@ -0,0 +1,41 @@
...
+      console.log('## Error: user metadata not found')
+      return
+    }
+    const { nick } = data.user.metadata
+
+    const filename: string = imageInfo.filename
+    e.preventDefault()
+    if (window.confirm('Are you sure?')) {
+      await actions.media.removeImage(imageInfo.mediaId, nick)
+      await removePhoto(filename)
```
vnugent: I would call removePhoto() to remove the photo from the backend first.  If it's successful then proceed to remove it from the local store.

src/components/media/RemoveImage.tsx
```diff
@@ -0,0 +1,41 @@
...
+      await toMyProfile()
```
vnugent: The full screen mode is still active in the parent gallery component when calling `toMyProfile()` so it's loading a different photo (the current index has changed after a removal).  I'd remove it and implement a callback as suggested to notify parent component.

vnugent (changes requested):

PR #451, opened 2022-07-31
fix #445

### LINKED ISSUE #445: Unable to start app locally with docker compose
To reproduce:
1. `git checkout develop`
2. `sudo docker-compose up`

Error encountered
image

As mentioned by @vnugent it is likely 
https://github.com/OpenBeta/open-tacos/commit/4910eb46538dcf0968cd222d1d41d5bdbae3e66c#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3
that fixed the tests but prevents docker-compose from running.

Removing those lines locally allows docker-compose up to start but run with errors, but removing said lines would prevent docker-compose workflow smoke test from running.

dlants:
Fixes https://github.com/OpenBeta/open-tacos/issues/445

I did some digging into this, and I think something's creating a recursive loop due to these interpolations in the compose config file:

```
      - SIRV_CLIENT_SECRET_RO=$${SIRV_CLIENT_SECRET_RO}
      - SIRV_CLIENT_SECRET_RW=$${SIRV_CLIENT_SECRET_RW}
      - NEXTAUTH_SECRET=$${NEXTAUTH_SECRET}
      - AUTH0_CLIENT_SECRET=$${AUTH0_CLIENT_SECRET}
      - AUTH0_MGMT_CLIENT_SECRET=$${AUTH0_MGMT_CLIENT_SECRET}
```

(That were added in https://github.com/OpenBeta/open-tacos/commit/4910eb46538dcf0968cd222d1d41d5bdbae3e66c#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3 ).

I think the double `$$` means that these are not interpolated by docker-compose, but instead are escaped (so the internal env is getting something like `SIRV_CLIENT_SECRET_RO=${SIRV_CLIENT_SECRET_RO}` See that mentioned here: https://docs.docker.com/compose/environment-variables/

next.js has some automagic variable expansion when it tries to apply the env. See here: https://nextjs.org/docs/basic-features/environment-variables

I think when trying to load the env, it takes the values passed in through docker-compose, it gets into an infinite loop trying to substitute them.

I'm not sure what the intent of adding these interpolated variables to that PR was. It seems like there's already a mechanism for passing these values into the container (the .env file), and it shouldn't be needed to also add them to the docker-compose file. Also, not sure why they were doubly-escaped.

### DIFF
--- a/.github/workflows/lighthouse-on-pr.yml
+++ b/.github/workflows/lighthouse-on-pr.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [pull_request]
+on: [push]
 jobs:
   lhci:
     name: Lighthouse
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,6 +1,6 @@
 name: 'tests'
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,9 +1,10 @@
-name: 'tests'
+name: 'CI'
 
 on: [push, pull_request]
 
 jobs:
   build:
+    name: Test build
     runs-on: 'ubuntu-latest'
     environment: "Preview"
     steps:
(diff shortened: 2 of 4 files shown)


vnugent:
Thanks for the fix.  I can run docker-compose locally.
I also made a few changes to skip e2e tests and the final build step on Pull request due to env variable access restriction.

Issue #462, opened 2022-08-01
Footer in user profile page not at bottom of screen when user has less photos.


xavier-kong:
The footer could be at the bottom of the page rather than close to the center.

The u page 
https://github.com/OpenBeta/open-tacos/blob/develop/src/pages/u/%5B...slug%5D.tsx 

calls the Layout component 
https://github.com/OpenBeta/open-tacos/blob/develop/src/components/layout.tsx 

which calls the Footer component 
https://github.com/OpenBeta/open-tacos/blob/develop/src/components/Footer.tsx

Reference image: the footer appears to be close to the center of the page. Could be an issue with handling users with few photos in profile. 
![image](https://user-images.githubusercontent.com/86543341/182157003-b0f225f7-6d8a-45b4-a3dd-118143351953.png)

Hendriksn:
Hello! Is this still an issue? I would like to work on that!

zichongkao:
Does not appear to be a problem anymore:
<img width="1263" alt="image" src="https://user-images.githubusercontent.com/3641356/235063382-cecb9418-a167-45d8-8db5-4cb50edbd212.png">


Can re-open if it still is!

PR #463, opened 2022-08-01
added fix for issue Footer in user profile page not at bottom of screen when user has less photos. #462

xavier-kong:
Tested when user has more photos than page can fit without scrolling and when user has few photos.
image


vnugent:
`h-screen` paints the white background only to the end of the screen.  When you scroll down, the rest is in black.

See https://open-tacos-c4rmdmvwz-openbeta-dev.vercel.app/?v=newTags

Screen Shot 2022-08-01 at 3 46 02 PM

xavier-kong:
I am using hard override dark mode on my machine so I did not catch that!

src/components/layout.tsx
```diff
@@ -12,7 +12,7 @@ interface LayoutProps {
 
 function Layout ({
   contentContainerClass = 'content-fullscreen-tablet',
-  rootContainerClass = 'root-container-default',
+  rootContainerClass = 'root-container-default flex flex-col',
```
vnugent, line 15: how about adding `flex flex-col` to `root-container-default` class in 
https://github.com/OpenBeta/open-tacos/blob/user-page-footer/src/styles/defaults.css#L16

vnugent (changes requested):

vnugent:
@xavier-kong can you have a look at this?

PR #467, opened 2022-08-02
Experiment with a global collections store

vnugent:


Downster (approved):
looks like the actual actions are empty and you just want to test and see how this works. If thats the case it looks good to me. I am not merging this incase you wanted to do something else but have left it approved for you.

vnugent:
I'll need to add additional skeleton actions and some unit tests

vnugent:
Obsolete.  We're now storing ticks in the db.

PR #487, opened 2022-08-11
Force page rebuild after an image deletion

vnugent:


vnugent:
Moved delete button to gallery

Screen Shot 2022-08-12 at 7 05 20 AM

Issue #480, opened 2022-08-04
Each main feed photo should be in a card


vnugent:
[New Tags view](https://openbeta.io/?v=newTags)
On desktop we can hover over an image to click on either the tag or the image's author link.  On mobile we will need to put each photo in a card with header and footer (similar to FB, IG, Twitter feed) in order for users to interact with photos.

### To do: 
- [x] Create a generic photo card component with header, body, footer (in ./src/components/ui dir) (https://github.com/OpenBeta/open-tacos/issues/518)
- [ ] Wrap each photo in the new card component
- [ ] We can add actions/links to header and footer in a separate PR

<img width="311" alt="Screen Shot 2022-08-04 at 9 52 16 PM" src="https://user-images.githubusercontent.com/3805254/182940739-1ecad3a3-22c8-480c-8fdd-43c87c068a8c.png">

osaf96:
@vnugent  I'm gonna look at this issue🙂

siman4457:
@osaf96 Any updates on this? I'm willing to help or pick this up if needed.

PR #502, opened 2022-08-17
Media convienience proposal

CocoisBuggy:
- May upload batches of images with a simple retry-recover
- Delete photo function appears in slideshow for auth'd users
- Delete photo function produced radix alert rather than window API
- Delete photo trigger is red, with trash icon
- Corners in gallery slightly rounded
- Fixed dropzone clicking not opening explorer in Chrome (tested on ubuntu)


vnugent:
thanks for this.  In the original PR by @amandril, delete button was supported in the slideshow view but due to the convoluted way that we try to support back button while fighting NextJ SSR, I decided to make the delete button only available in gallery view.   May I suggest that we fix bugs in this PR and address delete button in a separate PR?

edited:
In future releases we can adopt NextJS new [nested layout feature](https://nextjs.org/blog/layouts-rfc) to consolidate `/u/<uid>` and `/p/photoId` pages.  Then the back button issue is longer an issue and we can add back the delete button in Slide viewer.

CocoisBuggy:
> I decided to make the delete button only available in gallery view

Could I potentially leave a message in the slide view for the time being?  

**"Enable power mode to delete images"** or something of the kind; I was only able to figure out how to delete images by browsing the github issues - which I don't think the average user will know to do

vnugent:
> Could I potentially leave a message in the slide view for the time being?
> 
> **"Enable power mode to delete images"** or something of the kind; I was only able to figure out how to delete images by browsing the github issues - which I don't think the average user will know to do

Yes, please. That should work.  

On mobile (which we're aiming to focus on) once #480 is done, we can add a context menu on the header (similar to IG, Twitter post) where photo can be flagged or deleted.

vnugent:
I just did some testing on the Preview build.  Currently in production when I upload a photo from the main page or any other it used to 
1.  Disable the button to prevent another upload
2.  Take me to the user profile page

it's no longer the case.  Can you check?

current uploading in progress

Screen Shot 2022-08-20 at 11 35 51 AM

PR #503, opened 2022-08-20
Added PWA configs

vnugent:
#493
Still needs a lot of tuning


vnugent:
We now have a mobile app

PR #505, opened 2022-08-20
Media Misc 2

CocoisBuggy:
Dialogues styled solely with tailwind, function mostly from headless ui and radix

image

Actually, some other stuff also pulled in here.
Tailwind palette has been expanded so that the band colors have more shades

image

image

image

(Palettes generated on https://uicolors.app/create)


tailwind.config.js
```diff
@@ -8,17 +8,53 @@ module.exports = {
...
     extend: {
       fontFamily: {
         sans: ['ISO', ...defaultTheme.fontFamily.sans]
       },
       colors: {
         'custom-green': '#8BB191',
-        'ob-primary': '#F15E40',
-        'ob-secondary': '#8BB191',
-        'ob-tertiary': '#62cae3',
+        'ob-primary': {
```
vnugent: I'd merge this as is but I'd hate to see custom colors being removed later now that we have adopted Daisy theme.  They recommend only [2 additional shades per color](https://daisyui.com/docs/colors/)
- I'd define 2 more shades for OB brand's primary color: See `accent` in https://github.com/OpenBeta/open-tacos/blob/develop/tailwind.config.js#L47

vnugent (changes requested):

tailwind.config.js
CocoisBuggy: If the shades present a style risk I'll strip them out, but they are completely opt-in (The DEFAULT specifies that `bg-ob-primary` is its own color, seperate from `bg-ob-primary-[shade]`)

I'll strip the shades out, they aren't actually needed for this PR

Issue #346, opened 2022-06-08
names can be mangled in the breadcrumbs


atstp:
The [1-6-9 Boulder](https://tacos.openbeta.io/crag/dce84fc1-77f2-5760-80f0-6ac44d7d878e) is renamed to the "6-9 Boulder" in the breadcrumbs. 

[`sanitizeName`](https://github.com/OpenBeta/open-tacos/blob/a6292d268c9feb46aac1bde998f04cc720246c46/src/js/utils.ts#L98) seems to be the culprit

![1-6-9 Boulder becomes 6-9 Boulder](https://user-images.githubusercontent.com/5891072/172662951-f470c85c-69f7-495e-b334-bc8e78faedcf.png)

vnugent:
Good catch!  We need to a systematic clean up effort on the data side to remove index from area names, `1.` `a.` , etc

PR #508, opened 2022-08-22
- Prevent text mangling as screen size changes

CocoisBuggy:
- Shrunk slightly
- Icon as shown in figma docs
- When relevant, current item has text highlight

Resolution to  #346

### DIFF
--- a/src/components/ui/BreadCrumbs.tsx
+++ b/src/components/ui/BreadCrumbs.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import Link from 'next/link'
 import { sanitizeName } from '../../js/utils'
+import { LocationMarkerIcon } from '@heroicons/react/outline'
 /**
  * Turn each element of `pathTokens` to a gatsby-link.
  *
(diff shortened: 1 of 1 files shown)


src/components/ui/BreadCrumbs.tsx
```diff
@@ -22,27 +22,36 @@ interface BreakCrumbsProps {
 
 function BreadCrumbs ({ pathTokens, ancestors, isClimbPage = false }: BreakCrumbsProps): JSX.Element {
   return (
-    <div aria-label='area-breadcrumbs'>
+    <div aria-label='area-breadcrumbs' className='flex-wrap flex gap-2 text-sm items-center'>
+      <svg xmlns='http://www.w3.org/2000/svg' className='h-5 w-5 text-ob-primary' fill='none' viewBox='0 0 24 24' stroke='currentColor' strokeWidth={2}>
```
vnugent: Is this the location marker?  Would it be better to use `<LocationIcon />` from the @heroicon icon package (already part of the project dep) to improve readability/semantic?

CocoisBuggy: Good catch, I'll fix that

Issue #376, opened 2022-06-20
Updating user profile raises spurious "Username is already taken" after user has already claimed username


CocoisBuggy:
1. Go to user profile
2. Change username
3. hit save
4. change name or bio
5. hit save

You will get the error. Somewhere the form is trying to reclaim the username even though it wasn't modified and the user already owns it.

This blocks users from changing their name or bio

PR #509, opened 2022-08-22
Visual clarity + Fixed profile editing

CocoisBuggy:
- added some small documentation
- updateUserMetadata now works, previously would not commit if old nick had not been changed

fix for #376, documentation is in code but there is the critical discovery I made
```ts
// The auth0 management client only touches properties that
// are actually specified in metadata. So an undefined property
// is passed over, not nulled, by Auth0
const currentMeta = await getUserMetadata()
if (currentMeta.nick === metadata.nick) {
  delete currentMeta.nick
}

// Actually write this new data into the users profile.
const user = await auth0ManagementClient.updateUserMetadata(
  { id },
  metadata
)
```

Solution evidenced by https://github.com/auth0/node-auth0/blob/2e09f288cd6fc385e699a163e4c10b8692666d91/src/RetryRestClient.js#L53 and https://auth0.com/docs/manage-users/user-accounts/user-profiles/root-attributes/update-root-attributes-for-users Which suggest that patch will not touch props not defined in the object (must be explicitly set to null for deletion)

### DIFF
--- a/src/components/ui/TextField.tsx
+++ b/src/components/ui/TextField.tsx
@@ -2,13 +2,19 @@ import React from 'react'
 import { Field, FieldMetaProps } from 'formik'
 
 interface TextFieldProps {
+  /** key name for data */
   name: string
+  /** HTML label rendered for this input */
   label: string
+  /** Is this text input multiple lines? */
   multiline?: boolean
+  /** (If multiline) how many rows should this input have */
   rows?: number
   spellcheck?: boolean
+  /** Wait until submit to validate, or perform the validation immediately */
   validateImmediately?: boolean
-  validate?: (value: any) => Promise<undefined|string> // return an error message or undefined for valid input
+  /** return an error message or undefined for valid input */
+  validate?: (value: any) => Promise<undefined|string>
 }
 
 interface FieldType {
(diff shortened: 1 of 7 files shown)


vnugent:
I did a quick test in the Preview build.  I have to click Save twice for changes to write to Auth0.   Can you check?

vnugent:
Another note:  I think #376 has to do with my buggy logic checking Formik dirty/submitted flag.  If we can wait (I don't think the issue is a P1/critical.) to migrate to react-hook-form in the near future, it'd be a lot simpler to deal with form states.

CocoisBuggy:
> Another note: I think #376 has to do with my buggy logic checking Formik dirty/submitted flag. If we can wait (I don't think the issue is a P1/critical.) to migrate to react-hook-form in the near future, it'd be a lot simpler to deal with form states.

My fix here was to change the `/api/user/profile` to behave the way I expected it to

If you look in the [metadata client source](https://github.com/auth0/node-auth0/blob/2e09f288cd6fc385e699a163e4c10b8692666d91/src/RetryRestClient.js#L53) you can see that the metadata client feeds data up to the path call. The specified [behavior of Auth0 patch call](https://auth0.com/docs/manage-users/user-accounts/user-profiles/root-attributes/update-root-attributes-for-users](https://auth0.com/docs/manage-users/user-accounts/user-profiles/root-attributes/update-root-attributes-for-users#removing-attributes)) indicates that supplied params get re-written. Unless I've missed a step in the flow of data from form to back-end this is where the error is coming from (If it is Formik, then I'm not sure why my fix works)

You may be right that going away from Formik will fix the issue - but I'd still like to provide the intended behavior for the API endpoint (If a user updates their nickname to be the same nickname (I.E does not change it), we shouldn't allow an error to feed back as that is not a violable usage?)

If I've got it wrong I'll drop it from the PR, but I thought it would be more robust to modify the logic of the endpoint than the logic of the form (The form is bound to take many shapes in the lifetime of our usage of Auth0)

> I did a quick test in the Preview build. I have to click Save twice for changes to write to Auth0. Can you check?

Good catch, not sure how I didn't see this in testing

vnugent:
> My fix here was to change the /api/user/profile to behave the way I expected it to

If you see a bug there and this PR fixes it and users can save profile with 1 click (instead of 2) then we can merge this.

vnugent:
The issue still exists in the latest Preview build  https://open-tacos-fi2m9a55w-openbeta-dev.vercel.app
Steps to reproduce:
1.  Go to your profile page.  
2. Update your user name to something else
3. Hit Save.   (I can confirm username has changed in Auth0).
4. Now go to the Bio box and change some text.  The form validation logic runs again and it complains: 'User name is already taken'

CocoisBuggy:
Man this is one hydra of a form! This is thankfully just old state not getting updated when the request goes through

vnugent:
This may not be 100% related (Sorry if I'm stating the obivous).  When you smoke test your build locally, it's always good to do a complete build with `yarn build && yarn start`.  Because `yarn dev` has code hot reloading enabled.

vnugent:
Verified.  Thanks for your patience!

Issue #474, opened 2022-08-04
Set up a new Dev tenant on our Image hosting provider


vnugent:
We need to make it easier for new contributors to work on the code base (currently they need to ask for Sirv API keys for their local dev env). 

### To do
Configure a separate Dev tenant on Sirv.com, our image hosting provider so that we can leave Read-only API key in `.env` config 

- [x] Email viet at openbeta.io for new Dev keys
- [x] Update [.env file](https://github.com/OpenBeta/open-tacos/blob/develop/.env#L5-L8)
- [x] Build open-tacos locally and make sure you can upload photos
- [x] Submit a pull request and make sure the CI deploys a working site with photos

PR #511, opened 2022-08-24
New dev tenant and built-in Read-Only api key

amandril:
Issue #474


vnugent:
FYI the Vercel build has failed as expected because Sirv tenant & API keys have changed.  I'll update build envs i GH actions & Vercel.

Issue #370, opened 2022-06-17
Add photo timestamp to mobile view


vnugent:
### Use case
As a user I want to know when a photo was taken.

For current behavior see  [my profile](https://tacos.openbeta.io/u/vietnguyen) in mobile view.

### Tips:
- UserMedia component: https://github.com/OpenBeta/open-tacos/blob/develop/src/components/media/UserMedia.tsx#L64
- Convert `imageInfo.ctime`  to Date
- Display it when isDesktop is false
- Use a modern date util such as [date-fns](https://date-fns.org/) to format and display photo timestamp in human friendly format.  I think "x months ago or 3 weeks ago" for anything less than a year is easier to understand than May 2022.  Your thoughts?

<img width="309" alt="Screen Shot 2022-06-17 at 9 24 34 PM" src="https://user-images.githubusercontent.com/3805254/174401112-79d2c642-39bc-44e3-ad49-767ec0de2249.png">

osaf96:
Hi, I would love to do this task. It would be first contribution and hope to learn and contribute more and more. please assign this to me. Thanks

osaf96:
Thanks

vnugent:
@osaf96 Unfortunately the gallery feature won't work for you in the dev env without Auth0 access key (see https://github.com/OpenBeta/open-tacos/issues/389).   Until we figure out a solution, would you see if there's another https://github.com/OpenBeta/open-tacos/labels/good%20first%20issue?

osaf96:
oh ok, Let me know when you find the solution and when I can work on it.

vnugent:
Hi @osaf96 
I'm just checking to see if you are still planning to work on this issue (I believe I have given you API keys via email)

siman4457:
@vnugent I can pick this up once I have the API keys.

vnugent:
> @vnugent I can pick this up once I have the API keys.
Thanks.  I've just sent you an email.

PR #512, opened 2022-08-30
Add photo timestamp to mobile view #370.

siman4457:
Timestamps are displayed like "2 hours ago".
  Any posts older than 1 year have timestamps that display the month and year.

dec 2010
2 hours ago

### DIFF
--- a/src/components/media/UserMedia.tsx
+++ b/src/components/media/UserMedia.tsx
@@ -46,12 +46,12 @@ export default function UserMedia ({ index, uid, imageInfo, onClick, tagList, on
   const loader = isDesktop ? DesktopPreviewLoader : MobileLoader
   const shareableUrl = `/p/${uid}/${basename(imageInfo.filename)}`
 
-  const getUploadDate = (dateUploaded) => {
-    dateUploaded = new Date(dateUploaded)
+  const getUploadDate = (dateUploadedStr: string): string => {
+    const dateUploaded = new Date(dateUploadedStr)
     const currentTime = new Date()
     if (differenceInYears(currentTime, dateUploaded) >= 1) {
       return format(dateUploaded, 'MMM yyyy')
-    }  
+    }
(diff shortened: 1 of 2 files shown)


vnugent (changes requested):
Can you revert changes to `yarn.lock` file?  I think some dependencies were upgraded (minor version) causing the build to fail.

siman4457:
Reverted yarn.lock

vnugent:
@all-contributors add @siman4457 for code

vnugent:
@siman4457 I didn't see linting errors earlier. Can you please fix them?  we should be good to merge after that.

src/components/media/UserMedia.tsx
```diff
@@ -45,6 +46,15 @@ export default function UserMedia ({ index, uid, imageInfo, onClick, tagList, on
   const loader = isDesktop ? DesktopPreviewLoader : MobileLoader
   const shareableUrl = `/p/${uid}/${basename(imageInfo.filename)}`
 
+  const getUploadDate = (dateUploadedStr: string): string => {
```
vnugent: The build is failing in Vercel ([see log](https://vercel.com/openbeta-dev/open-tacos/EDmc8cYWDeLHqKyBVALsisvmd3UZ)).  I believe `ctime` is already a Date object.  
> Type error: Argument of type 'Date' is not assignable to parameter of type 'string'.
07:45:02.493

FYI now that build env issues have been resolved if you do a normal `git commit` without --no-verify, the pre-commit hook will run a full build locally and can catch these types of issues.  
Thank you for your patience!

vnugent (changes requested):

src/components/media/UserMedia.tsx
siman4457: Oops, I missed that. Thanks for fixing the build env issues. I made an edit, ran the pre-commit hook, and the build was successful for me locally. However, I could not view this in the browser since I needed to remove the SIRV image hosting keys from my local.env.

PR #533, opened 2022-09-14
Fix _ is not defined error after lib upgrades

vnugent:
After PR #530 (probably due Nextjs lib upgrade to library upgrade)  search broke with error `_ is not defined`

### DIFF
--- a/src/components/search/XSearch.tsx
+++ b/src/components/search/XSearch.tsx
@@ -20,13 +20,13 @@ export default function XSearch ({ isMobile = true, placeholder = 'Climb search'
       isMobile={isMobile}
       placeholder={placeholder}
       getSources={async ({ query }) => {
-        const sources = [...await xsearchTypesense(query), await searchPoi(query)]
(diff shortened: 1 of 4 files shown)


src/components/search/XSearch.tsx
```diff
@@ -20,13 +20,13 @@ export default function XSearch ({ isMobile = true, placeholder = 'Climb search'
       isMobile={isMobile}
       placeholder={placeholder}
       getSources={async ({ query }) => {
-        const sources = [...await xsearchTypesense(query), await searchPoi(query)]
-        // This may look a little bizarre, but it's just so that the sources appear
-        // in the same order that we render them (climbs, poi, areas, fa)
-        // If there's a better unified way, I'd love to know how - Coco
-        return [sources[0], sources[3], sources[1], sources[2]]
+        const sources = await xsearchTypesense(query)
```
vnugent, line 23: `xsearchTypesense()` returns an array so we simply return it as is.  They're destructured by their source id in render().

 fyi  @CocoisBuggy

PR #544, opened 2022-09-22
518: Create a Twitter/IG style card component

siman4457:
Created a general card component. Moved some logic to utils.

### DIFF
--- a/src/components/ui/Card/Card.tsx
+++ b/src/components/ui/Card/Card.tsx
@@ -7,11 +7,11 @@ export interface CardProps {
   styles?: string
 }
 
-export default function Card({
+export default function Card ({
   header,
   image,
   body,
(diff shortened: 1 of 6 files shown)


src/components/ui/Card/Card.tsx
```diff
@@ -0,0 +1,55 @@
...
+  header: string | JSX.Element
+  body: string | JSX.Element
+  styles?: string
+}
+
+export default function Card({
+  header,
+  image,
+  body,
+  styles = 'bg-white drop-shadow rounded-lg'
```
vnugent: Can you please use  the following Daisy-UI classes (so that we can support additional themes in the future)? Thanks!
- bg-white --> `bg-base-100` 
- rounded-lg --> `rounded-box`

vnugent (changes requested):

Issue #405, opened 2022-07-05
Display backer list 


vnugent:
Show the current donor list, amount raised to date.

- [x] Ping viet at openbeta dot io for OpenCollective API key
- [x] Put API key in .env.local
- [x] Write a component that fetches donor list (see example below)
- [x] Add a new tab to the main page (see screenshot)

https://medium.com/open-collective/open-collective-graphql-api-preview-3b42ed1d55ff

<img width="750" alt="Screen Shot 2022-07-04 at 6 47 18 PM" src="https://user-images.githubusercontent.com/3805254/177325748-fb01a2c9-f6d1-46fe-9d1b-92876b9f0019.png">

PR #545, opened 2022-09-24
Add financial backers page OpenBeta#405

admanny:
Related to #405 
image


vnugent:
@admanny thank you very much!

Issue #547, opened 2022-09-25
Remove duplicate financial donors


vnugent:
### Issue
It seems like when someone contributes multiple times to different tiers, their data is shown for each tier.
- Go to https://openbeta.io/?v=backers
- Adelore and Alex appear 2x 

### To do

In [FinancialBackers.tsx](https://github.com/OpenBeta/open-tacos/blob/develop/src/components/home/FinancialBackers.tsx#L15) remove duplicates using **underscore** [uniq](https://underscorejs.org/#uniq) (lib already included in the project) or something equivalent.

Theakayuki:
Created fix with underscore like recommended, and setup pull request.

siman4457:
UI could also use an update on mobile. Seems to be too many cards per row and too much text per card. Consider reducing font size on mobile. ![image](https://user-images.githubusercontent.com/28658492/192810960-0c38a16d-fd41-44a7-b624-b5e84423696f.png)

vnugent:
@siman4457 can you please create an issue?

PR #548, opened 2022-09-26
Updated FinancialBackers to remove Duplicate donor issue #547

Theakayuki:
#547 Using underscore Uniq added a Quick filter to take out duplicate names, would recommend that we have some kind of account id in case you run into the names being the same. Didn't know if that was coming from another entry point on the backend could only follow the typing back to the index and didn't find where it was being populated to add ids further up.

### DIFF
--- a/src/components/home/FinancialBackers.tsx
+++ b/src/components/home/FinancialBackers.tsx
@@ -6,12 +6,8 @@ export interface FinancialBackersProps {
   donors: FinancialBackerAccountType[]
   totalRaised: string
 }
-
 export default function FinancialBackers({ donors, totalRaised }: FinancialBackersProps): JSX.Element {
-
-
-  const uniqDonors = uniq(donors, (donor) => donor.account.name)  // Using the underscore library to get unique donors, based on their name
-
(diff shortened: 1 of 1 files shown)


vnugent:
> would recommend that we have some kind of account id in case you run into the names being the same

Thanks for the PR!  You've raised a good point I haven't thought about.   To get account id we need to update the GraphQL query  and update `FinancialBackerAccountType` Typescript:
- Insert a new line with `id` above `name`: https://github.com/OpenBeta/open-tacos/blob/develop/src/pages/index.tsx#L228
- Declare id type:  `id: string` https://github.com/OpenBeta/open-tacos/blob/develop/src/js/types.ts#L269

vnugent:
@all-contributors add @Theakayuki for code

vnugent:
@Theakayuki can you run `yarn lint` locally and fix the linting error(s)?  Then we're good to merge.  We can address the sort by account ID in a separate PR.  Thanks!

```
FinancialBackers.tsx:9:41: Missing space before function parentheses. (@typescript-eslint/space-before-function-paren)

```

Theakayuki:
Gotcha went ahead and ran it should be all good now.

PR #557, opened 2022-09-30
View user ticks integrated along with re-import ticks button for mtn …

Downster:
…project ticks


vnugent:
@Downster I merged the delete ticks PR and I think it's caused conflicts. Can you fix them? we should be good to merge after. Thanks

Downster:
@vnugent So the delete-ticks PR actually had all of the changes that were made in this branch. I created the delete-ticks branch from this version.

PR #560, opened 2022-09-30
feat: Use financial donor account id to ensure unique

### LINKED ISSUE #555: Use financial donor account id to ensure unique
Related issue #547 

Currently we're using name as a key to ensure uniqueness.  Short term fix: https://github.com/OpenBeta/open-tacos/pull/548

A better approach would be grouping the donor list by account id.

### Todo
- Insert a new line with id above name: https://github.com/OpenBeta/open-tacos/blob/develop/src/pages/index.tsx#L228
- Declare id type: id: string https://github.com/OpenBeta/open-tacos/blob/develop/src/js/types.ts#L269
- update `uniq` function to use id instead of name https://github.com/OpenBeta/open-tacos/blob/develop/src/components/home/FinancialBackers.tsx#L10

```javascript
  const uniqDonors = uniq(donors, (donor) => donor.account.id) // Using the underscore library to get unique donors, based on their name
```

Balastrong:
This PR fixes #555.

### DIFF
--- a/src/components/home/FinancialBackers.tsx
+++ b/src/components/home/FinancialBackers.tsx
@@ -7,7 +7,7 @@ export interface FinancialBackersProps {
   totalRaised: string
 }
 export default function FinancialBackers ({ donors, totalRaised }: FinancialBackersProps): JSX.Element {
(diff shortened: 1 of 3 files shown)


vnugent:
@all-contributors add @Balastrong for code

PR #573, opened 2022-10-09
Adding photo upload error alert when upload fails (fixes #436)

### LINKED ISSUE #436: User friendly message when upload image too large
Display a user friendly message when upload image is too large (or when there's anything wrong with the upload)

### To do:
- [x] Write a simple popup alert based on https://www.radix-ui.com/docs/primitives/components/alert-dialog (we're phasing out `headless-ui`
- [x] handle client side error:  https://github.com/OpenBeta/open-tacos/blob/develop/src/js/hooks/usePhotoUploader.ts#L36
- [x] handle server side error: https://github.com/OpenBeta/open-tacos/blob/develop/src/js/hooks/usePhotoUploader.ts#L36

clintonlunn:
fixes #436

Taking a first stab at this. 

Added a basic alert to tell the user that their upload failed. 

The message will say either:
 - 'Failed to upload: Exceeded retry limit'.
 - 'Failed to upload: max file size 5MB.'

image

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -3,6 +3,8 @@ import { useRouter } from 'next/router'
 import MobileTabletAppBar from './MobileAppBar'
 import DesktopAppBar from './DesktopAppBar'
 import useResponsive from '../js/hooks/useResponsive'
+import PhotoUploadError from './media/PhotoUploadError'
+import { userMediaStore } from '../js/stores/media'
 
 const NAV_BAR_IDENTIFIER = 'tacos-nav-bar'
 
@@ -13,6 +15,8 @@ interface HeaderProps {
 export default function Header (props: HeaderProps): JSX.Element {
   const { isTablet, isMobile } = useResponsive()
   const includeFilters = Boolean(props.showFilterBar)
+  const photoUploadErrorMessage = userMediaStore.use.photoUploadErrorMessage()
+  const isPhotoError = photoUploadErrorMessage !== null
 
   const router = useRouter()
   const isIndexPage = router.pathname === '/'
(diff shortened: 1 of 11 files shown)


clintonlunn:
I found another ErrorAlert component that I was able to reuse. To my understanding, this doesn't use the headless-ui like issue said is getting phased out. 
image

image

The mobile view doesn't look awesome, but neither does the rest of the page behind it.

vnugent:
@clintonlunn Thanks for this.
You're right, you're phasing out headless-ui.  It looks like there isn't enough room for new import ticks button.  We can fix that in a separate issue.

clintonlunn:
@vnugent Are we okay with the pop-up being in the middle of the screen on both mobile and desktop? It seems like we have both middle and slide up from bottom (when on mobile). I've been trying to figure out how to slide from and stick to bottom but I am not super familiar with how that could be done using tailwind.

vnugent:
> Are we okay with the pop-up being in the middle of the screen on both mobile and desktop?

Yes please.

clintonlunn:
> > Are we okay with the pop-up being in the middle of the screen on both mobile and desktop?
> 
> Yes please.

Great! then I will leave it how it is. Do you see an opportunity to expand test coverage here? I'm mulling over what sort of functionality we can test.

vnugent:
> Great! then I will leave it how it is. Do you see an opportunity to expand test coverage here? I'm mulling over what sort of functionality we can test.

If you move the error popup logic downstream to UserGallery.tsx, it may be easier to test since there is already a [test](https://github.com/OpenBeta/open-tacos/blob/develop/src/components/media/__tests__/UserGallery.tsx).

Then you can mock 'userMediaStore.use.photoUploadErrorMessage()' to return true and check for the the popup.

clintonlunn:
@vnugent just updating. moving things to UserGallery was pretty straight forward. however, having trouble mocking that `userMediaStore.use.photoUploadErrorMessage()`

clintonlunn:
Actually, after some additional testing it looks like we might want to expand where the error popup could occur. Since the user can upload a photo from essentially all of the website (since there is a "+ Photo" button in nav bar), it might make sense to put this photo error logic into the `Header` component. It seems to work fine so far.

vnugent:
You're right.  Another good place would be the parent of Header, the application root [_app.tsx](https://github.com/OpenBeta/open-tacos/blob/develop/src/pages/_app.tsx).  You can see there's already a similar mechanism for Toast.

vnugent:
I spent a bit of time on this to see if we can mock/set the error message in test.  Unfortunately, _app.tsx is called by NextJS build, not an ordinary component that we can render in test. 

see https://github.com/vercel/next.js/discussions/13717

we can live without a unit test for now.

clintonlunn:
@vnugent thanks for looking into this!

vnugent:
My apologies.  I keep flip flopping on where to house the popup logic.  
I think Header is actually a better option like you original suggested since we can easy mock that component in test. If you can move the  popup logic to Header I'd be happy add a test.

clintonlunn:
@vnugent ok should be where you want now. I agree, having it at the tippy top _app level is probably not appropriate. Are you going to push a test to this branch? i'm wondering what i was missing in being able to mock that store value to trigger the photo error.

vnugent:
I've pushed a test:  https://github.com/OpenBeta/open-tacos/pull/573/files#diff-2b48f67fd6cccf60de91b72f2dc3af02e71bcce989643496ac1bae42d715513b

Thanks for your help.  I think we're good to merge

vnugent:
Screen Shot 2022-10-22 at 3 07 41 PM

Issue #592, opened 2022-11-03
Upgrade Next to 13


vnugent:
Need to upgrade hero icons first because the new next image lib requires alt tag which existing version of heroicons doesn't have.

- [x] Upgrade heroicons 
- [ ] upgrade next

mikeschen:
Closing, we have upgraded to Next 13.

PR #594, opened 2022-11-04
Upgrade next to v13

vnugent:
Issue #592

Upgrade Next v12 -> v13

- [x] Fix image component
- [x] Fix Link component
- [ ] Build issue


vnugent:
Postponed.

As much as I'd like to incrementally adopt the new v13 [nested routing](https://beta.nextjs.org/docs/routing/fundamentals) feature (main motivation for upgrading), it's not worth the effort to fix all the build issues that come with the upgrade.

Issue #599, opened 2022-11-06
Shorten latitude & longitude


vnugent:
### Issue
Latitude & longitude are shown with all of their significant digits.

Example:
44.36890442356387,  -121.13926266934996
https://openbeta.io/areas/f3c655b6-9c95-597f-aaa6-fd7d8a4865ec

### To do

Use `toFix(5)` ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed)) to shorten lat & lng in
- src/pages/areas/[id].tsx
- src/pages/crag/[id].tsx

<img width="240" alt="Screen Shot 2022-11-06 at 7 58 48 PM" src="https://user-images.githubusercontent.com/3805254/200189786-715d9d21-3b77-4a15-9495-d58e760a2287.png">

christophermorin:
Hello. I'd like to take a crack at this. I'll send a PR up, hopefully it's what you are looking for.

PR #600, opened 2022-11-07
fix: shorten displayed lat and long coords

christophermorin:
#599 

Screenshot 2022-11-06 at 21-29-15 Navajolands • Cove Area Area • 8 climbs
Screenshot 2022-11-06 at 21-30-49 Northern Arizona • Navajolands Area • 65 climbs

### DIFF
--- a/src/components/area/panel/panelHeader.tsx
+++ b/src/components/area/panel/panelHeader.tsx
@@ -50,7 +50,7 @@ export function PanelHeader (props: PanelHeaderProps): JSX.Element {
             text-sm'
             title='Click to view on google maps'
           >
-            {props.latitude}, {props.longitude}
+            {props.latitude.toFixed(5)}, {props.longitude.toFixed(5)}
           </div>
         </a>
       </div>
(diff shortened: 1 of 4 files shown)


vnugent:
@all-contributors add @christophermorin for code !

vnugent:
@christophermorin Thank you so much for this fix.
Sorry I have left incorrect instructions in the original issue (#599) .  Since we pass lat,long to a URL to google maps, we want to preserve all precisions if possible, only shorten the display values.

 Instead of shorten the props in the parent components, can you apply toFix() to 2 places where they're rendered?  Thank you!

- https://github.com/OpenBeta/open-tacos/blob/develop/src/components/area/panel/panelHeader.tsx#L53 
- https://github.com/OpenBeta/open-tacos/blob/develop/src/components/crag/cragSummary.tsx#L158

christophermorin:
@vnugent Should be displaying correctly now. Though I did notice a small bug.
Currently when you are sending the coordinates to google, it is doing so backwards. 
Latitude is going into the search field as longitude and vise versa. The cords for Alabama lead to Antarctica at the moment :)

Just to clarify it seems to only happen with /area pages, /crag goes to the correct coordinates.

Issue #499, opened 2022-08-14
Photos uploading from a route page should be tagged with that route by default


ygingras:
If I uploaded a photo from my profile, I don't expect any default tags, but if I uploading by following the link from a route page, I almost certainly mean to upload a photo of that route.

CocoisBuggy:
I'll work on this, it seems like a reasonable behavior assumption for that button

CocoisBuggy:
I have a fix for this that is halfway there

There is now a button that explicitly uploads-and-tags. I'll bring up the behavior of the top button at this weeks meeting and we can discuss how it should work.

Either way, this is a good piece of functionality to have started work on.

![image](https://user-images.githubusercontent.com/64557383/185947639-99aef45c-837d-4403-90da-21389d33b851.png)

#507 Will add this

vnugent:
Fixed in PR #619 

Instead of adding a new Upload button, I added additional logic to the current button to determine the current page from Next useRouter() hook and add a tag accordingly.

PR #619, opened 2022-11-26
Auto tag newly uploaded photo from area or climb page

vnugent:
Issue #499 
Instead of adding a new Upload button, I added additional logic to the current button to determine the current page from Next `useRouter()` hook and add a tag accordingly.

Also signal Next to rebuild the page being tagged.


src/js/hooks/usePhotoUploader.ts
```diff
@@ -47,20 +47,13 @@ export default function usePhotoUploader ({ onUploaded }: UploaderProps): PhotoU
       imageData = Buffer.from(imageData)
     }
 
-    let retry = 3
```
vnugent, line 50: For some reason this causes photos to be uploaded 3x in dev mode

Issue #620, opened 2022-11-30
Can't upload photos from iphone due to image size restriction


vnugent:
Currently users can only upload a photo up to 5MB.  Photos taken from modern mobile phone can be 7-8MB.

PR #621, opened 2022-11-30
Raise file size limit and apply reduction

vnugent:
Issue #620


vnugent:
This approach won't work due to Next placing a limit on serverless functions.
https://vercel.com/guides/how-to-bypass-vercel-body-size-limit-serverless-functions

Issue #629, opened 2022-12-19
Area description only shows 1st paragraph


vnugent:
When an area description contains multiple paragraphs, only the the 1st paragraph is shown.

https://openbeta.io/areas/740bf011-c032-5f86-987c-270009a89267

If you click on Edit (required an account) you will see the rest of the description.

clintonlunn:
Completed in https://github.com/OpenBeta/open-tacos/pull/639

PR #639, opened 2022-12-29
Adding read more / read less behavior to areas (#629)

clintonlunn:
I repurposed the Description from the crag summary components since it had a read more/less behavior.

@vnugent Is this approach of using transitions of maxheight to show/hide text instead of removing it from the dom preferred? I suppose this approach gives us the ability to have smooth transitions where other approaches wouldn't.

### DIFF
--- a/src/components/area/panel/panelHeader.tsx
+++ b/src/components/area/panel/panelHeader.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react'
-import { useMemo } from 'react'
-import { summarize } from '../../crag/cragSummary'
+import Description from '../../ui/Description'
 
-export interface PanelHeaderProps{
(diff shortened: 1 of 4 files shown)


clintonlunn:
I cannot find any areas that have description in the environment that Vercel deploys the preview in.

clintonlunn:
https://open-tacos-git-area-description-readmore-629-openbeta-dev.vercel.app/areas/58343ac7-d09d-502d-80d2-8c6694f7c914

image
image
image
image

vnugent:
Thanks for refactoring the Description component. 
> Is this approach of using transitions of maxheight to show/hide text instead of removing it from the dom preferred? I suppose this approach gives us the ability to have smooth transitions where other approaches wouldn't.

I think it's fine.

Issue #626, opened 2022-12-14
User Profile Edit: Website url validation is too strict


evannoronha:
## Problem

I'm trying to add a link to my instagram account to my OpenBeta profile. My handle is `@evanclimbs.jpg`. The corresponding url is `https://www.instagram.com/evanclimbs.jpg/`.

## Current Behavior
<img width="442" alt="image" src="https://user-images.githubusercontent.com/6803522/207706881-5dd0ec0d-d899-4cd6-b4be-e4bf87aa0d28.png">

## Possible Solution
Use a standard url validation library to check this field.

## Steps to Reproduce
1. Visit `https://openbeta.io/account/edit`
2. Input a valid url under `Website` that contains dots in the path.
3. Observe failure.

vnugent:
@evannoronha Thanks for reporting.   I did try Yup built-in URL validator but it was too strict.  

### Suggestion
Improve the current regex https://github.com/OpenBeta/open-tacos/blob/develop/src/js/utils.ts#L173 or find a better validator

vnugent:
@evannoronha the issue has been fixed. Can you please try updating your profile again?

vnugent:
Verified in prod

PR #648, opened 2023-01-07
Coco/url validation

CocoisBuggy:
fix for #626

This makes the url validation for flexible, and closes a validation hole in the `/api/user/profile` to check the website before pumping it to the auth0 datastore.

**explicit http** is also prohibited (https being the enforced protocol)

I've added cases into the test file, if I've missed something obvious changing the validator function will enforce both on profile and on api endpoint.


vnugent:
@CocoisBuggy thanks for this!

vnugent:
@all-contributors add @evannoronha for bug

PR #683, opened 2023-02-03
Address Overflow Issue

### LINKED ISSUE #678: Overflow issue user gallery page
This issue tracker is **ONLY** used for reporting bugs!

Horizontal scroll bar is visible when the screen size is below 1022px, Appears to be caused by long image captions.
http://localhost:3000/u/viet-test-profile

## Expected Behavior
Horizontal scrollbar should be hidden on small screen sizes.

## Possible Solution
add `overflow: hidden` to image container

## Steps to Reproduce
1. Navigate to user gallery page: http://localhost:3000/u/viet-test-profile
2. Make screen smaller than 1022px
3. Horizontal scroll bar should be visible. See Screen Shot.

Screen Shot 2023-02-01 at 1 38 35 PM

Image with `overflow: hidden` added to  the image container `<div class="relative"></dvi>`

Screen Shot 2023-02-01 at 1 38 43 PM

tedgeving:
Set a width on `Tag.tsx` component to make the text truncate properly.  
https://github.com/OpenBeta/open-tacos/issues/678

### DIFF
--- a/src/components/media/Tag.tsx
+++ b/src/components/media/Tag.tsx
@@ -9,10 +9,9 @@ interface PhotoTagProps {
   isAuthorized?: boolean
   showDelete?: boolean
   size?: 'md' | 'lg'
-  maxWidth?: number
 }
 
-export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, isAuthorized = false, maxWidth }: PhotoTagProps): JSX.Element | null {
+export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, isAuthorized = false }: PhotoTagProps): JSX.Element | null {
   const [url, name] = resolver(tag)
   if (url == null || name == null) return null
   const isArea = tag.destType === TagTargetType.area
--- a/src/components/media/Tag.tsx
+++ b/src/components/media/Tag.tsx
@@ -1,8 +1,8 @@
 import Link from 'next/link'
 import { XCircleIcon, MapPinIcon } from '@heroicons/react/20/solid'
 import clx from 'classnames'
-
 import { HybridMediaTag, MediaTagWithArea, MediaTagWithClimb, TagTargetType } from '../../js/types'
+import useResponsive from '../../js/hooks/useResponsive'
 
 interface PhotoTagProps {
   tag: HybridMediaTag
@@ -16,6 +16,8 @@ export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, i
   const [url, name] = resolver(tag)
   if (url == null || name == null) return null
   const isArea = tag.destType === TagTargetType.area
+  const currentScreensize = useResponsive()
+
   return (
     <Link href={url} prefetch={false}>
       <a
@@ -27,7 +29,7 @@ export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, i
         onClick={stopPropagation}
       >
         {isArea && <MapPinIcon className='w-4 h-4' />}
-        <div className='mt-0.5 whitespace-nowrap truncate text-sm'>{name}</div>
+        <div style={currentScreensize.isDesktop || currentScreensize.isTablet ? { maxWidth: 240 } : { maxWidth: 'calc(80vw)' }} className='mt-0.5 whitespace-nowrap truncate text-sm'>{name}</div>
         {isAuthorized && showDelete &&
           <button onClick={(e) => {
             onDelete(tag.id)
--- a/src/components/media/Tag.tsx
+++ b/src/components/media/Tag.tsx
@@ -1,5 +1,7 @@
 import Link from 'next/link'
-import { XCircleIcon, MapPinIcon } from '@heroicons/react/20/solid'
+import { XCircleIcon } from '@heroicons/react/20/solid'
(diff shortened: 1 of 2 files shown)


vnugent:
###  Gallery view on mobile 
we use an entirely different card component for mobile: https://github.com/OpenBeta/open-tacos/blob/develop/src/components/media/MobileMediaCard.tsx#L19

Screen Shot 2023-02-04 at 9 31 19 AM

---
### Signed in / desktop
We should account for the X button when a user signs in and views their own profile

Screen Shot 2023-02-04 at 9 31 54 AM

tedgeving:
@vnugent Let me know if those changes work for you. Normally I'd add a few css rules to sort this kind of issues out, or let me know if you think I should try a different approach.

src/components/media/Tag.tsx
```diff
@@ -27,7 +30,8 @@ export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, i
         onClick={stopPropagation}
       >
         {isArea && <MapPinIcon className='w-4 h-4' />}
-        <div className='mt-0.5 whitespace-nowrap truncate text-sm'>{name}</div>
+
+        <div style={maxWidth !== undefined ? { maxWidth: maxWidth } : { maxWidth: DEFAULT_TAG_WIDTH }} className='mt-0.5 whitespace-nowrap truncate text-sm'>{name}</div>
```
vnugent: what do you think if we just default max width to Tailwind `max-width-xs` which is 320?  And set `title` attribute to `name` to show a tool tip on desktop?  This way we don't have to introduce the maxWidth param?
```html
<div className='mt-0.5 whitespace-nowrap truncate text-sm max-w-xs ' title={name}>{name}</div>
```

https://tailwindcss.com/docs/max-width

src/components/media/Tag.tsx
```diff
@@ -10,12 +9,16 @@ interface PhotoTagProps {
   isAuthorized?: boolean
   showDelete?: boolean
   size?: 'md' | 'lg'
+  maxWidth?: number
 }
 
-export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, isAuthorized = false }: PhotoTagProps): JSX.Element | null {
+export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, isAuthorized = false, maxWidth }: PhotoTagProps): JSX.Element | null {
```
vnugent: with `max-width-xs` set we don't have to introduce maxWidth param

vnugent (changes requested):

src/components/media/Tag.tsx
tedgeving: I looked into using that class an it is not narrow enough for some screen sizes and the small cards. 240px or so is the max width for several screen sizes and layouts.  See Screen Shot.

Maybe it would be better to limit the tag length on the admin side?  Or alternatively write a small utility function that truncates long tags?

Screen Shot 2023-02-08 at 7 22 07 PM

vnugent: Ah, you're right. In which case we can specify an arbitrary value directly in Tailwind: 

`max-w-[240px]`

This way we can keep all css within className (as much as possible).  The title attribute will enable a tool tip for those on desktop.

tedgeving: LOL, that was easy, first projects I've used tailwind on.  Nice feature, I've used Bootstrap for 10 plus years and it does not have anything similar.

Should be good to merge now.

src/components/media/Tag.tsx
```diff
@@ -10,12 +9,16 @@ interface PhotoTagProps {
   isAuthorized?: boolean
   showDelete?: boolean
   size?: 'md' | 'lg'
+  maxWidth?: number
```
vnugent: can you remove this?

src/components/media/Tag.tsx
```diff
@@ -10,12 +9,16 @@ interface PhotoTagProps {
...
 }
 
-export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, isAuthorized = false }: PhotoTagProps): JSX.Element | null {
+export default function Tag ({ tag, onDelete, size = 'md', showDelete = false, isAuthorized = false, maxWidth }: PhotoTagProps): JSX.Element | null {
   const [url, name] = resolver(tag)
   if (url == null || name == null) return null
   const isArea = tag.destType === TagTargetType.area
+
+  const DEFAULT_TAG_WIDTH = 320
```
vnugent: can you remove?

src/components/media/Tag.tsx
vnugent: please remove `maxWidth` prop

vnugent (changes requested):

src/components/media/Tag.tsx
tedgeving: done

src/components/media/Tag.tsx
tedgeving: done

vnugent:
Screen Shot 2023-02-10 at 1 40 57 PM
Screen Shot 2023-02-10 at 1 40 00 PM

Issue #688, opened 2023-02-08
Clicking on user name should take me to their profile


vnugent:
## Expected Behavior
When I click on a user name or the user icon I expect to go to their user profile

## Current Behavior
Instead the link takes me to the tagged page

## Possible Solution
Look at RecentImageCard component https://github.com/OpenBeta/open-tacos/blob/develop/src/components/home/RecentMediaCard.tsx#L30

## Steps to Reproduce
Go to https://openbeta.io

---
<img width="885" alt="Screen Shot 2023-02-08 at 2 32 20 PM" src="https://user-images.githubusercontent.com/3805254/217548265-bfeb18db-c784-4bd4-bcf7-535f4e851c20.png">

ParthParolekar:
I would like to work on this issue. Can you assign this to me?

PR #690, opened 2023-02-08
fix: added position relative to the anchor tag containing the post image

ParthParolekar:
Fix for #688 

The `loaded` component in `RecentMediaCard.tsx` has a `position : absolute` and takes the entire width and height of its first parent element with a position of relative.

 `<div className={clx('absolute top-0 left-0 w-full h-full', loaded ? 'bg-transparent' : 'bg-gray-50 bg-opacity-60 border animate-pulse')}>{loaded}</div>`

The first parent element with `position: relative` was the `<Card/>` component with the `card` class so the `loaded` component takes up the entire height and width of the card which is why clicking on the username takes the user to the tagged page.
image

The fix was as simple as adding a `position : relative` to the anchor/link tag containing the image.

### DIFF
--- a/src/components/home/RecentMediaCard.tsx
+++ b/src/components/home/RecentMediaCard.tsx
@@ -19,7 +19,10 @@ interface RecentImageCardProps {
 /**
  * Image card for the home page
  */
-export const RecentImageCard = ({ imageInfo, tagList }: RecentImageCardProps): JSX.Element => {
+export const RecentImageCard = ({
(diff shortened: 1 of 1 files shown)


vnugent:
@ParthParolekar Thank you.  Sometimes a good fix doesn't mean having to write a lot of code :D

vnugent:
@all-contributors add @ParthParolekar for code

Issue #662, opened 2023-01-24
Support changing area left-to-right order


vnugent:
### Use case
By default areas are shown by alphabetical order.  For many climbing areas it makes sense to list crags by their physical location, for example, as you follow the access trail.

<img width="895" alt="Screen Shot 2023-01-24 at 10 13 21 PM" src="https://user-images.githubusercontent.com/3805254/214417991-eaa86291-7844-4f83-bea9-74bc2b6ebe8c.png">

---

### Context
We do have code for drag-n-drop'ing climbs built by another contributor a while back.  It's currently not in use to due to recent refactoring of the climb and area page layout to support the editing feature.

Crag Table (drag n drop to reorder climbs - currently not in use)
https://github.com/OpenBeta/open-tacos/blob/develop/src/components/crag/cragTable.tsx

Library (already added to package.json)
https://github.com/atlassian/react-beautiful-dnd

### Suggestions

Unlike its cousin, the climb object, area doesn't have a `leftToRightIndex` field that can be used for sorting.  I can help adding it to the backend.

I'd add the sorting and drag-n-drop code to AreaCRUD component:
https://github.com/OpenBeta/open-tacos/blob/develop/src/components/edit/AreaCRUD.tsx#L43

tedgeving:
I'd be willing to take a look a this one if it is still open.

vnugent:
Thanks @tedgeving.  I've assigned it to you and added additional context to the issue.  I can help with the backend portion.

tedgeving:
@vnugent 
this commit is a WIP  the drag and drop component is working.  
1. Need some assistance adding `leftToRightIndex ` to the component data on the backend.
2. What component should handle saving the newly ordered routes? Little confused on how that works in the app?

vnugent:
I created https://github.com/OpenBeta/openbeta-graphql/issues/221 to track the backend work.  I'll see if someone can help or I'll look into doing it myself.

Can you open a PR with what you current have? (label it WIP) I'll comment on the code.

A rough outline for the next steps:

1.  Add the new field (TBD), `leftToRight` to the GQL query that feeds the data to the crag page
https://github.com/OpenBeta/open-tacos/blob/develop/src/pages/crag/%5Bid%5D.tsx#L156
2.  Update the corresponding Typescript declarations
3. Make sure we pass the new prop all the way down to `cragSummary` component and `AreaCRUD`.
4.Sort the list based on the new leftToRight prop
[...]

PR #696, opened 2023-02-09
Issue 662 area order and sorting

tedgeving:
Work In Progress Addresses part of the following Issue
https://github.com/OpenBeta/open-tacos/issues/662

DND Library has been (https://github.com/atlassian/react-beautiful-dnd) added to the Area  edit page
https://github.com/OpenBeta/open-tacos/blob/develop/src/components/edit/AreaCRUD.tsx#L43


vnugent:
Nice!  Can you check the preview build?  The list is showing up twice
https://open-tacos-git-issue-662-area-order-openbeta-dev.vercel.app/crag/793d57d9-1f16-5457-8810-cc4e05e07421

tedgeving:
@vnugent Removed the duplicate listing.

vnugent:
@tedgeving what do you think if we merge this PR now and create a follow up one when the backend functionality is done?

tedgeving:
That works for me.

tedgeving:
@vnugent that works for me.

vnugent:
@tedgeving any chance you can revive this PR?

If you can rebase the PR with `develop` and get the drag-n-drop UI working again, I can finish the backend integration part.

tedgeving:
Let me take a look at it, should not be a problem.

Thanks for keeping me posted, have been busy with the day job and chasing
the good snow we've had here in NM.  Hoping to get to contributing
again soon.

glassbead0:
I love this feature or ordering areas! Just tracking down the current progress and I see the backend work is done, and looks like the frontend is still a WIP.

I'm curious if it makes sense to implement the area sorting UI using the same CSV Power Editor that is used for route sorting (cut/paste areas in the order you want), instead of a drag and drop. While I think drag and drop is just fine, it does create inconsistency. I have limited time to help with this (and no experience yet with the tech stack), but if I do get time I will start looking at code to see if I can help at all.

zichongkao:
@vnugent Two outstanding problems: 

1. In testing, I'm hitting the duplicate index error for `leftRightIndex`. I think it's because I'm sending `[{areaId: .., leftRightIndex: 0}, {areaId: .., leftRightIndex: 1}, .., {areaId: .., leftRightIndex: <# of childareas -1>}]` and when we try to set the leftRightIndex to be 0 for the first item, there is already another entry in the db with leftRightIndex 0, so throws the error even though later on, that item's leftRightIndex will replaced by a different entry. I could reset everything to -1 and then reassign, but it's not that elegant. Curious if you have a better approach in mind.
2. The drag and drop works fine in a single column, but you can't drag things across two columns. It doesn't detect the x coordinate, only the y coordinate, so it thinks you're dropping it into the leftmost column. Basically I think react-beautiful-dnd only supports either horizontal or vertical lists. When we have two columns, we need both directions.
Screen Recording 2023-05-17 at 4 02 09 PM

tedgeving:
The previous version was only one column in the edit mode to avoid this issue IIRC… On May 17, 2023, at 4:05 PM, zkao ***@***.***> wrote:﻿
@vnugent Two outstanding problems:

In testing, I'm hitting the duplicate index error for leftRightIndex. I think it's because I'm sending [{areaId: .., leftRightIndex: 0}, {areaId: .., leftRightIndex: 1}, .., {areaId: .., leftRightIndex: <# of childareas -1>}] and when we try to set the leftRightIndex to be 0 for the first item, there is already another entry in the db with leftRightIndex 0, so throws the error even though later on, that item's leftRightIndex will replaced by a different entry. I could reset everything to -1 and then reassign, but it's not that elegant. Curious if you have a better approach in mind.
The drag and drop works fine in a single column, but you can't drag things across two columns. It doesn't detect the x coordinate, only the y coordinate, so it thinks you're dropping it into the leftmost column.

vnugent:
Unfortunately, wrapped list isn't supported by the DnD library.  See this long discussion: https://github.com/atlassian/react-beautiful-dnd/issues/316

An alternative? https://github.com/clauderic/dnd-kit

zichongkao:
Yup -- that's what I landed on too. Seems pretty mature and well-supported. Now we can handle multiple columns:
Screen Recording 2023-05-18 at 1 06 41 AM

zichongkao:
Next up, we'd have to update `src/components/crag/cragTable.tsx` to do the same there.

zichongkao:
A good place to play with the functionality is https://open-tacos-git-issue-662-area-order-openbeta-dev.vercel.app/crag/f3c655b6-9c95-597f-aaa6-fd7d8a4865ec

src/components/crag/cragSummary.tsx
```diff
@@ -117,7 +117,7 @@ export default function CragSummary ({ area, history }: CragSummaryProps): JSX.E
    * Initially set to the childAreas prop coming from Next build, the cache
    * may be updated by the users in the AreaCRUD component.
    */
-  const [childAreasCache, setChildAreasCache] = useState(childAreas)
+  const [childAreasCache, setChildAreasCache] = useState(sortByLeftRightIndex(childAreas))
```
zichongkao, line 120: One question worth discussing is whether we should be sorting in the frontend or backend. You could argue that frontend is better if we want to allow dynamic sorting (L->R, Alphabetical, Most children) one day, so the backend result remain invariant and can be cached. Backend sorting just simplifies frontend code a bunch especially in the case where we always sort L->R. 

I realize that for areas, L->R sorting doesn't make too much sense because areas aren't lined up on a wall like climbs. So the main purpose of the sorting here is just a stable ordering. If stability is what we care about, then perhaps no point enabling multiple sort orders for areas, which would lead to backend sorting being better.

For now I'm just following the design pattern of climbs, and doing sorting on the frontend.

vnugent: All good points.  Agreed we should sort in the backend.  Can you create a backend issue?

src/components/edit/AreaCRUD.tsx
```diff
@@ -4,9 +4,28 @@ import { AreaSummaryType } from '../crag/cragSummary'
 import { DeleteAreaTrigger, AddAreaTrigger, AddAreaTriggerButtonMd, AddAreaTriggerButtonSm, DeleteAreaTriggerButtonSm } from './Triggers'
 import { AreaEntityIcon } from '../EntityIcons'
 import NetworkSquareIcon from '../../assets/icons/network-square-icon.svg'
+import React, { useState } from 'react'
+import useUpdateAreasCmd from '../../js/hooks/useUpdateAreasCmd'
+import { useSession } from 'next-auth/react'
```
vnugent: Not super important. I usually place library imports at the top and then our own stuff.  `ts-standard` doesn't enforce this rule but I think it helps you quickly gain some context when scanning the import section.

https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

package.json
```diff
@@ -7,6 +7,9 @@
     "@algolia/autocomplete-js": "1.7.1",
     "@algolia/autocomplete-theme-classic": "1.7.1",
     "@apollo/client": "^3.6.9",
+    "@dnd-kit/core": "^6.0.8",
```
vnugent, line 10: Is the old `react-beautiful-dnd` still being used elsewhere?  if not can you yarn remove it?

zichongkao: Yeah it's still being used for CragTable. After this, I'll switch CragTable over and we remove `react-beautiful-dnd`.

src/components/crag/cragSummary.tsx
zichongkao: https://github.com/OpenBeta/openbeta-graphql/issues/294 Sounds good. I can work on this next.

src/components/edit/AreaCRUD.tsx
zichongkao: Fixed!

package.json
vnugent: This crag table?  I don't think we use this component anywhere. 
https://github.com/OpenBeta/open-tacos/blob/develop/src/components/crag/cragTable.tsx

zichongkao: Seems like it's used in https://openbeta.io/finder. I didn't realize we had a page like that. Doesn't seem fully functional, but seems like a useful page. What do you think we should do with it? Fix it up? Deprecate the page?

vnugent: I believe there are 2 crag table components. Anyway, I'm merging this now and we can clean it up in a follow up PR.

zichongkao: Follow up here: https://github.com/OpenBeta/open-tacos/pull/839

Issue #701, opened 2023-02-14
Discord invite invalid


vnugent:
## Expected Behavior

Discord invite link should let me join Discord, instead it says: "Invite Invalid"

<img width="552" alt="Screen Shot 2023-02-14 at 8 30 01 PM" src="https://user-images.githubusercontent.com/3805254/218838469-62e76bde-3ee9-4d3a-bffe-1ad16c7bfa7a.png">

## Current Behavior
Error:   Invite Invalid

## Possible Solution
Update all expired links
1.  In your IDE do a file search for: "discord.gg".  You should see 6 pages with links to Discord
2. Replace the old link with: `https://discord.gg/ptpnWWNkJx` 

## Steps to Reproduce
The link is shown in various places including the README file.
- Non-logged in users: https://openbeta.io
- Logged in:  Profile -> Discord

ramzyraz:
Hi @vnugent ,
I would like to work on this issue. Can you assign this to me.

PR #702, opened 2023-02-15
fix: Updated the expired Discord invite links

ramzyraz:
Issue #701


vnugent:
Thank you for your help!
@all-contributors add @ramzyraz for code

PR #703, opened 2023-02-15
Fix `updateMyProfile` bug that is blocking users from updating profiles without specifying a website

diericx:
If a user does not enter a website (for example, if they are a new user attempting to simply edit their username) they will get an error.

I think this utility function is confusing and should either be renamed to betray the fact that it is not just a check, or it should be changed to a simple check with the correction logic being moved elsewhere. 

I also think the unnecessary recursion should be removed.

These are out of scope though. This PR simply fixes the bug that is affecting users without altering the `checkWebsiteUrl` function at all.

A secondary issue is that these specific error messages are not being propagated to the front end. That would have made debugging easier, and would have given me a bit more feedback as a normal user as to what was going on.

### DIFF
--- a/src/pages/api/user/profile.ts
+++ b/src/pages/api/user/profile.ts
@@ -37,7 +37,7 @@ const updateMyProfile: Handler = async (req, res) => {
 
     // validate that URL is properly formatted if one is provided
     let website = req.body?.website
-    if (website !== null && website !== "") {
+    if (website != null && website !== '') {
       website = checkWebsiteUrl(req.body?.website)
       if (website == null) {
         throw new Error('Bad website URL')
--- a/src/pages/api/user/profile.ts
+++ b/src/pages/api/user/profile.ts
@@ -34,9 +34,16 @@ const updateMyProfile: Handler = async (req, res) => {
     if (!checkUsername(req.body?.nick)) {
       throw new Error('Bad username')
     }
-    // validate that URL is properly formatted
-    const website = checkWebsiteUrl(req.body?.website)
-    if (website == null) { throw new Error('Bad website URL') }
+
+    // validate that URL is properly formatted if one is provided
+    let website = req.body?.website
+    if (website !== null) {
+      website = checkWebsiteUrl(req.body?.website)
+      if (website == null) {
+        throw new Error('Bad website URL')
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


vnugent:
I tried to clear out the website box and hit save.  The error was: Bad website URL.
You can checkout the preview build here: https://open-tacos-git-fork-diericx-fix-profile-patch-openbeta-dev.vercel.app/

Screen Shot 2023-02-15 at 3 43 20 PM

vnugent:
> I think this utility function is confusing and should either be renamed to betray the fact that it is not just a check, or it should be changed to a simple check with the correction logic being moved elsewhere.
> I also think the unnecessary recursion should be removed.
> These are out of scope though. This PR simply fixes the bug that is affecting users without altering the checkWebsiteUrl function at all.
> A secondary issue is that these specific error messages are not being propagated to the front end. That would have made debugging easier, and would have given me a bit more feedback as a normal user as to what was going on.

The profile edit form was built on older form components (buggy) built on top of Formik.  Newer form components use react-hook-form which I think is easier to work with.  We just didn't have the bandwidth to migrate the profile to react-hook-form.

diericx:
Ahh I believe there is one final change I forgot to push... req.body.website is coming in as empty string not null. I can push it when I get home in a few hours sorry about that!

edit: attempting from my phone

src/pages/api/user/profile.ts
```diff
@@ -34,9 +34,16 @@ const updateMyProfile: Handler = async (req, res) => {
     if (!checkUsername(req.body?.nick)) {
       throw new Error('Bad username')
     }
-    // validate that URL is properly formatted
-    const website = checkWebsiteUrl(req.body?.website)
-    if (website == null) { throw new Error('Bad website URL') }
+
+    // validate that URL is properly formatted if one is provided
+    let website = req.body?.website
+    if (website !== null && website !== "") {
```
vnugent: I think issue might be here.  When website is `undefined`  website !== null will evaluate to true.  

`website != null` is a better test.

diericx: Ahh okay. Will wait until getting to my laptop to check it over and make another commit then.

vnugent (approved):

vnugent:
Thanks so much for your contribution and (should have been a quick merge.  Sorry I totally missed the notification on my feed)

@all-contributors add @diericx for ideas and code

Issue #591, opened 2022-11-02
Investigate infinite scroll 


vnugent:
### Use case

I want to scroll through the main feed to see recent photo uploads.  Currently the list is hard-coded at around 30 items.

vnugent:
Fixed in https://github.com/OpenBeta/open-tacos/pull/869

PR #707, opened 2023-02-16
feat: Implement Infinite-scroll on user profile

kunal00000:
issue: #591 

Initial number of photos -> 20
number of photos added on scroll -> 9
max number of photos -> 100 

Screenshot 2023-02-16 at 10 57 06 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "react-dropzone": "^14.2.1",
     "react-hook-form": "^7.34.2",
     "react-hotkeys-hook": "^3.4.7",
+    "react-icons": "^4.7.1",
     "react-map-gl": "^7.0.10",
     "react-paginate": "^8.1.3",
     "react-responsive": "^9.0.0-beta.6",
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "react-dropzone": "^14.2.1",
     "react-hook-form": "^7.34.2",
     "react-hotkeys-hook": "^3.4.7",
-    "react-icons": "^4.7.1",
     "react-infinite-scroll-component": "^6.1.0",
     "react-map-gl": "^7.0.10",
     "react-paginate": "^8.1.3",
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "react-hook-form": "^7.34.2",
     "react-hotkeys-hook": "^3.4.7",
     "react-icons": "^4.7.1",
+    "react-infinite-scroll-component": "^6.1.0",
     "react-map-gl": "^7.0.10",
     "react-paginate": "^8.1.3",
     "react-responsive": "^9.0.0-beta.6",
--- a/src/components/media/UserGallery.tsx
+++ b/src/components/media/UserGallery.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useState, useRef, Dispatch, SetStateAction, useEffect } from 'react'
 import { Dictionary } from 'underscore'
-import { TagIcon, ArrowPathIcon } from '@heroicons/react/24/outline'
+import { TagIcon } from '@heroicons/react/24/outline'
 import { useRouter } from 'next/router'
 import { basename } from 'path'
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -54,6 +54,7 @@
     "react-dropzone": "^14.2.1",
     "react-hook-form": "^7.34.2",
     "react-hotkeys-hook": "^3.4.7",
+    "react-icons": "^4.7.1",
```
vnugent: Is there a specific icon you're looking for?  we already have heroicon package installed

src/components/media/UserGallery.tsx
```diff
@@ -120,6 +121,34 @@ export default function UserGallery ({ uid, postId: initialPostId, auth, userPro
...
+  useEffect(() => {
+    // add new items from the imageList to the imageListToShow array when length changes
+    if (imageList !== undefined) setImageListToShow(imageList.slice(0, imageListToShowLength))
+    console.log('Length Before ', imageListToShow.length)
+  }, [imageList, imageListToShowLength])
+
+  useEffect(() => {
+    console.log('Length After ', imageListToShowLength)
+
+    const handleScroll = (): void => {
```
vnugent: Sorry I forgot to mention in our chat on Discord that we're currently using `react-use` library that provides a wealth of useful hooks.   You can use the `useIntersection()` hook (built on Intersection Observer API) to simplify the scroll detection code.  It'd be more performant and less custom code :)

- Docs: https://streamich.github.io/react-use/?path=/story/sensors-useintersection--docs
- StickyHeader component https://github.com/OpenBeta/open-tacos/blob/develop/src/components/crag/StickyHeader.tsx

In StickyHeader variable `atTop` becomes true when the `<div ref={intersectionRef} ...>` reaches the top of the screen.   You can change `rootMargin` to be notified when you reach the bottom of the screen (I don't remember exactly how on top of my head).

vnugent (changes requested):

package.json
kunal00000: I needed a Loader icon which I didn't find in heroicons. I checked it before using the react icons library.

vnugent: what about hero icon `arrow-path` with Tailwind `animate-spin` to create the loading effect?

kunal00000: Screenshot 2023-02-17 at 3 54 44 PM

It looks good. Let me revert the last commit.

src/components/media/UserGallery.tsx
kunal00000: I tried this but not working, maybe I'll have to dive deeper into the working of `useIntersection()` hook (learning `react`).
But, I solved the problem of custom code. I implemented this again using `react-infinite-scroll-component` library. It was quite simple to use that which I didn't expect before.

vnugent:
@kunal00000 
Thanks for the update.  Can you view the Preview build https://open-tacos-git-fork-kunal00000-infinite-scroll-openbeta-dev.vercel.app/u/viet-test-profile1 and let me know if the gallery is jittering or is it just my computer?

https://user-images.githubusercontent.com/3805254/220174994-0f1d9992-0450-4e82-92aa-8c981824adc3.mov

package.json
```diff
@@ -54,6 +54,8 @@
...
```
vnugent: Please yarn remove this package.

kunal00000:
@vnugent , Thanks for reviewing, 
Yes, it is jittering. would figure out the problem and update accordingly. 
If you have some suggestions please let me know.

kunal00000:
The reason behind jittering is `rotate` property in `animate-spin`. 
Corners of icon intersect with the viewBox because of which height of viewBox increase and decrease on loading.

https://user-images.githubusercontent.com/92316166/220189113-2e54f93a-8286-423b-a131-21482265478f.mov

----
What else we can do is instead of loader an down arrow with `animate-bounce`.

https://user-images.githubusercontent.com/92316166/220189450-23a25971-1e75-40d9-94e5-d40f05b9667a.mov

PR #711, opened 2023-02-17
Single column area list

diericx:
Area list is now a single column, more like a table.

### DIFF
--- a/src/components/edit/AreaCRUD.tsx
+++ b/src/components/edit/AreaCRUD.tsx
@@ -3,9 +3,8 @@ import { AreaType } from '../../js/types'
 import { AreaSummaryType } from '../crag/cragSummary'
 import { DeleteAreaTrigger, AddAreaTrigger, AddAreaTriggerButtonMd, AddAreaTriggerButtonSm, DeleteAreaTriggerButtonSm } from './Triggers'
 import { AreaEntityIcon } from '../EntityIcons'
-import NetworkSquareIcon from '../../assets/icons/network-square-icon.svg'
 
-export type AreaCRUDProps = Pick<AreaType, 'uuid'|'areaName'> & {
+export type AreaCRUDProps = Pick<AreaType, 'uuid' | 'areaName'> & {
   childAreas: AreaSummaryType[]
   editMode: boolean
   onChange: () => void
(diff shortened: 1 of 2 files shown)


vnugent:
@diericx Thank you for your contribution.  I'm curious to hear why you think a single column is more readable on wide screen than a 2-column layout? (the current layout does turn into 1 column on smaller screens)

Also the area icons as it's part of a new work-in-progress visual cues to help differentiate areas/crags/boulders vs climbs/boulder problems.   Why do you think we should remove them?

In general I'd appreciate it if you show us a simple mockup of changes before committing to the work.

---

Screen Shot 2023-02-18 at 8 02 31 AM

---
Screen Shot 2023-02-18 at 8 06 31 AM

diericx:
@vnugent 

> In general I'd appreciate it if you show us a simple mockup of changes before committing to the work.

Sorry, this was such a small change it seemed easier to just mock up in code and open a PR!

> icons

I didn't see a matching icon for boulders/crags so I opted for the consistency of text. Right now you can't even tell the difference between a crag and a boulder. I would rather have the info there and possibly improve it with icons than omit the info completely.

And personally I thought icons for Areas and text for Crags/Boulders looked weird but we can also do that if that is a migration path towards icons you prefer.

> why you think a single column is more readable on wide screen than a 2-column layout?

I was having a hard time scanning for specific data in the tall rows of two columns. I can revert that. It's much easier to read without the large number and with smaller rows.

Screenshot 2023-02-18 at 08 57 28

diericx:
And one with less data, more variety. Doesn't look great without the alignment of a table element, but at least the info is there. Will be much nicer with a filter form at the top.

Screenshot 2023-02-18 at 09 00 35

diericx:
I'll close this for now, sounds like it is more appropriate to discus design/UI decisions in Discord or in an issue rather than on a PR. Rather just close this now than let it go stale.

vnugent:
I'd appreciate your contribution. However, it'd make our life easier if you first propose your idea in Discord or create an issue with mockups so we can have a discussion.  In general I'd gladly accept bug fixes and UI improvements even if they're not on the current target milestone if they don't significantly alter what is currently there.

Issue #681, opened 2023-02-02
Show local climbing organization on area/climb page


vnugent:
### Use case 1 - View LCO profile
As a climber I want to learn more about the local climbing organization (LCO) when viewing a climb or an area page.

### Use case 2 - Update LCO profile
As a LCO admin I want to add my organization (name, website, social media link, contact, how to report bad bolts, **how to make a donation**)

In the pilot (MVP) phase we will invite a couple LCOs and place their info on openbeta.io under the respective areas.  We should not invest too much time into building a full-blown and scaleable workflow at this stage, but rather work with participating LCOs to learn more about their needs as well as getting feedback from climbers.

### Proof of concept
- Boulder Canyon (Colorado) https://openbeta.io/crag/e519a674-a620-509c-9e86-a246f84a8e40
- Beacon Rock (Washington) https://openbeta.io/crag/a8c395e6-9bbd-53ab-ba22-43995142b470
---
Mockup - use case 1

A user can hover on the button to see more info.

<img width="1209" alt="lco-mockup" src="https://user-images.githubusercontent.com/3805254/224161994-9db947b9-abf0-4ad7-970e-398d0d91ea1c.png">

zichongkao:
Completed by https://github.com/orgs/OpenBeta/projects/6

PR #723, opened 2023-03-09
Organization abstraction proposal

zichongkao:
Issue #681


vnugent:
@zichongkao I'm merging this to make it more visible.  We can always revise it as we go.

vnugent:
@all-contributors add @zichongkao for code & ideas

Issue #721, opened 2023-03-08
Simplify issue template


vnugent:
### Use case
As a user I want to report an issue or suggest an enhancement.  The current template on is asking me to provide too much information.

Todo: 

- [ ] Propose a simpler issue template
- [ ] Update the template file https://github.com/OpenBeta/open-tacos/blob/develop/ISSUE_TEMPLATE.md

```
This issue tracker is **ONLY** used for reporting bugs!

<!--- Provide a general summary of the issue in the Title above -->

## Expected Behavior
<!--- Tell us what should happen -->

## Current Behavior
<!--- Tell us what happens instead of the expected behavior -->

## Possible Solution
<!--- Not obligatory, but suggest a fix/reason for the bug, -->

## Steps to Reproduce
<!--- Provide a link to a live example, or an unambiguous set of steps to -->
<!--- reproduce this bug. Include a screen shot if it's easier. -->
1.
2.
3.
4.

## Context (Environment)
<!--- How has this issue affected you? What are you trying to accomplish? -->
<!--- Providing context helps us come up with a solution that is most useful in the real world -->

<!--- Provide a general summary of the issue in the Title above -->

## Detailed Description
<!--- Provide a detailed description of the change or addition you are proposing -->

## Possible Implementation
<!--- Not obligatory, but suggest an idea for implementing addition or change -->

```

PR #728, opened 2023-03-14
Simplify github issue template

musoke:
#721 suggests that the issue template is asking for too many details.  As someone who filed a couple issues yesterday, I agree.

I have reduced the number of sections as some seemed redundant. 

I also reordered them so that it asks:

1. What the reporter wanted to do.
2. How the user tried to do that
3. What happened
4. What they expected to happen
5. Optional suggestions of causes and fixes.


vnugent:
@all-contributors add @musoke for ideas and code

PR #731, opened 2023-03-16
fix: Allow spaces in coordinates

### LINKED ISSUE #730: Overly strict coordinate parsing
## Context (Environment)

I have been adding coordinates to some areas.  I have noticed that coordinates with a space in them are not allowed.  This is strange, given that that this is the format displayed by opentacos, e.g.
image

## Steps to Reproduce

1. Navigate to an area, enter edit mode
2. Add coordinates with a space between the latitude and longitude, for example "48.26000, 2.67000".  

## Expected Behavior

The editor accepts the coordinates and lets the user save them.

## Current Behavior

The editor complains that coordinates are invalid.

image

Removing the space is sufficient to fix the error.

## Possible Solution

Coordinates with one or more spaces between the comma and longitude should be allowed.  The space introduces no ambiguity.

musoke:
The editor checks user-provided latitude and longitude string string for two numbers separated by a comma.  This excludes common but unambiguous ways of formatting such strings with additional spaces.

Update `LATLNG_PATTERN` to allow for spaces on either side of either number.

Fixes #730

### DIFF
--- a/src/components/edit/EditAreaForm.tsx
+++ b/src/components/edit/EditAreaForm.tsx
@@ -9,7 +9,7 @@ import { AreaType, AreaUpdatableFieldsType, RulesType } from '../../js/types'
 import { areaDesignationToForm, areaDesignationToDb, AreaTypeFormProp, AreaDesignationRadioGroup } from './form/AreaDesignationRadioGroup'
 import useUpdateAreasCmd from '../../js/hooks/useUpdateAreasCmd'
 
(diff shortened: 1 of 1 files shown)


musoke:
Note that this change allows spaces on either side on the numbers, whereas your suggested fix was for just the one space - I think such spaces common when copying and pasting coordinates, especially with the latitude and longitude are pasted one at a time.

I couldn't test locally (can't authenticate without API keys) and jest failed in the precommit hook (seems unrelated?).  I tested the regex in https://regex101.com/.

Thanks for the pointers for a humble physicist!

musoke:
Works in the vercel preview above

Issue #733, opened 2023-03-17
Importing ticks gives no visual feedback upon completion


Urik:
Importing ticks gives no visual feedback upon completion.

Importing ticks gives no feedback besides adding the ticks themselves. When trying the feature without ticks in my Mountain Project profile, I thought something went wrong and tried again.

## Steps to Reproduce
1. Go to profile
2. Click on import ticks
3. Use my profile (https://www.mountainproject.com/user/200417239/uri-b)

## Expected Behavior
Some sort of feedback should happen letting me know the operation was successful. 

## Current Behavior
The "import ticks" popup is closed but nothing else happens.

## Possible Solution
Let's use `react-toastify` to give users feedback.

Related issue: https://github.com/OpenBeta/open-tacos/issues/590

Urik:
Heads up I'm working on this issue, commenting here as I found no  way to assign myself to the issue.

Urik:
PR up for review
https://github.com/OpenBeta/open-tacos/pull/736

vnugent:
PR merged and deployed.  Thank you for your help!

PR #736, opened 2023-03-18
Firing a toast when importing ticks is successful

Urik:
# Firing a toast when importing ticks is successful

## Description
Related to issue https://github.com/OpenBeta/open-tacos/issues/733

Importing ticks used to give no feedback besides adding the ticks themselves. When trying the feature without ticks in my Mountain Project profile, I thought something went wrong and tried again.

This PR fixes the issue by adding a toast message of type info saying "<number> ticks have been imported!".

## Screenshot
image

### DIFF
--- a/src/components/users/ImportFromMtnProj.tsx
+++ b/src/components/users/ImportFromMtnProj.tsx
@@ -3,6 +3,7 @@ import { Transition } from '@headlessui/react'
 import { FolderArrowDownIcon, XMarkIcon } from '@heroicons/react/24/outline'
 import { useMutation } from '@apollo/client'
 import { useSession } from 'next-auth/react'
+import { toast } from 'react-toastify'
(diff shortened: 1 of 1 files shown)


src/components/users/ImportFromMtnProj.tsx
```diff
@@ -71,6 +72,9 @@ export function ImportFromMtnProj ({ isButton }: Props): JSX.Element {
             input: ticks
           }
         })
+
+        const ticksCount = ticks?.length ?? 0;
```
Urik: I've added this kind of sanitization just in case since I don't actually know what the API returns if an abnormal situation arises.  
If it's not needed let me know and I'll just use `ticks.length`

vnugent:
@all-contributors add @Urik for code

Issue #718, opened 2023-03-04
Username Validation Incorrect Warning


dan-kat-dev:
This issue tracker is **ONLY** used for reporting bugs!

## Expected Behavior

When editing your profile user name, you get an error when there are spaces in the user name. The warning message should probably say 'username cannot contain spaces' or something.

## Current Behavior

![image](https://user-images.githubusercontent.com/62715496/222917613-66356a00-506e-4dda-8970-5782370c8bd1.png)
The actual error message says 'Must start and end with a letter or number' which I think is incorrect

## Possible Solution

Show correct user name validation warning, maybe something like "User name cannot contain spaces" or "User name must contain only letters and numbers"

vnugent:
@daniel-n-kaye thank you for suggesting this!

@all-contributors add @daniel-n-kaye for idea

dan-kat-dev:
@vnugent I'd be willing to try to take this on as a first issue, if that's ok, as it seems like it would hopefully be a pretty isolated update.

vnugent:
The error messages are listed here

https://github.com/OpenBeta/open-tacos/blob/6f08aa18892fb7fbf730129030b40a2c67184e33/src/components/users/account/ProfileEditForm.tsx#L18

Since profile update code requires authentication with Auth0, please send an email: viet at openbeta dot io and I'll reply with some additional API keys for the dev environment.

shoaib-31:
@vnugent I think my pull request will do the work.

hawkishpolicy:
Is this still active?

clintonlunn:
This is still an active issue. @danekshea any updates on your changes?

vnugent:
I'm working on moving some of user profile functionality to our own DB.  Username update is the first piece so I can take on this issue.  What do you all think the error message should be?

Username validation rules:
1.  cannot contain spaces or special characters
2. must begin with a letter

See https://github.com/OpenBeta/open-tacos/issues/830

hawkishpolicy:
Based on my understanding of the regex, it appears that it requires a username to begin and end with a letter or number.

It cannot contain spaces but it can contain "." or "-" or "_" 

Would these be considered special characters?

I was brainstorming on how best to write this error message so that users can understand it clearly.

What about:

1. Must begin and end with a letter or number.
2. Can contain "." or "-" or "_"
3. Cannot contain spaces.

It's verbose but it is specific. 

@zichongkao had some ideas. 

[...]

PR #758, opened 2023-03-26
Fix: issue 718, improve validation message for user profile name edit

danekshea:
Quick fix for https://github.com/OpenBeta/open-tacos/issues/718 with an added description of the regex. Didn't test the build due to not having auth available, might be overflow due to length of message but the responsive UI probably handles it find.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "tailwindcss": "^3.1.8",
     "ts-standard": "11.0.0",
-    "typescript": "4.9.5"
+    "typescript": "4.4.3"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "tailwindcss": "^3.1.8",
     "ts-standard": "11.0.0",
-    "typescript": "4.4.3"
+    "typescript": "4.9.5"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


vnugent:
can you please smoke test the change using the preview build?   https://open-tacos-aa73c08rd-openbeta-dev.vercel.app/

thanks!

vnugent:
Screenshot 2023-04-07 at 5 54 32 AM

you're right - there's not enough room for the long error message.

zichongkao:
I looked into this briefly to find a path forward.

I see several options:
1. Change the error message display to accommodate larger messages. We could move to a toast, or a tool-tip like modal that is displayed on top of the page content. It could perhaps auto-disappear after several seconds or once the error is fixed.
2. Break up `checkUsername` (https://github.com/OpenBeta/open-tacos/blob/d9f2d1a3011f4dfefb4861718840440eca24f3a7/src/js/utils.ts#L163) into its three constituent subrules:
- "First and last characters have to be alphanumeric",
- "User name can only consist of alphanumeric and special characters: '.', '-', '_'"
- "At most one special character is permitted"
each with its own check and dedicated error message. 
3. Try a shorter error message: 'Username must be alphanumeric, optionally separated by a ".", "-" or "_".'
4. Simplify the `checkUsername` constraints so that it is more easily explained. Eg. "All alphanumeric only", or "Alphanumeric and special characters (., /, -)" + "special characters not permitted in first and last"  (ie remove the "at most one special char" rule)

I'd favor 3 and 4. I feel like the current logic is somewhat hard to explain and fairly arbitrary. Perhaps "Only alphanumeric characters and ".", "_", "-" are permitted" would do.

zichongkao:
@vnugent @danekshea What do y'all think?

vnugent:
I'd be in favor of 3 and/or 4.  Like you said we can break down the validation into multiple tests, each showing their own shorter message.

danekshea:
3 and 4 sound pretty logical @zichongkao !

vnugent:
superseded by https://github.com/OpenBeta/open-tacos/pull/838

PR #765, opened 2023-03-30
fix: bring back the crag edit history

### LINKED ISSUE #667: Bring back page edit history
### Use case
When viewing a climb/area page I want to see the edit history

This feature was already implemented and available in the old area page.  We just need to port it over.

Old area page:  https://openbeta.io/areas/18c5dd5c-8186-50b6-8a60-ae2948c548d1 (click on Edit -> Histor)
New ara page: https://openbeta.io/crag/18c5dd5c-8186-50b6-8a60-ae2948c548d1

saferthanhouses:
### Show ChangeHistory on the crag detail page

Screen Shot 2023-03-30 at 11 36 11 AM
Screen Shot 2023-03-30 at 11 36 00 AM

resolves OpenBeta/open-tacos#667


src/components/crag/cragSummary.tsx
```diff
@@ -205,6 +219,8 @@ export default function CragSummary (props: AreaType): JSX.Element {
       })
     }
     setCache({ ...formData })
+
+    await getAreaHistory()
```
vnugent: The query introduces additional delay to the Save action (button being disabled).

- Whenever we call the backend GQL to update an area or climb, we also trigger Next to rebuild the page.
https://github.com/OpenBeta/open-tacos/blob/develop/src/js/hooks/useUpdateAreasCmd.tsx#L67
- We can also decrease the [revalidation value](https://github.com/OpenBeta/open-tacos/blob/0ba7142bee501ff163f2d358240edf780ceb9bb9/src/pages/crag/%5Bid%5D.tsx#L203) to increase the page "freshness".  

If you really want to avoid showing stale history, we can delay the query execution by moving it to the history popup and only call the it when the user wants to view the history. Otherwise Next will update the page in 30s (or whatever the revalidation value is).

src/components/crag/cragSummary.tsx
```diff
@@ -298,8 +314,23 @@ export default function CragSummary (props: AreaType): JSX.Element {
...
-          <div className='mt-4 text-right' id='editTogglePlaceholder'>
-            <EditModeToggle onChange={setEditMode} />
+          <div className='mt-4 justify-end flex flex-row' id='editTogglePlaceholder'>
+            <div className='mr-2' id='editTogglePlaceholder'>
+              <EditModeToggle onChange={setEditMode} />
+            </div>
+            {!isEmpty(changeHistory) && (
+              <MobileDialog>
+                <DialogTrigger>
+                  <Button
```
vnugent: Can you please use `<button className='btn btn-secondary ...'>`?  
See https://daisyui.com/components/button/

We're deprecating the custom Button component.  Sorry this is not documented anywhere.

vnugent (changes requested):

src/components/crag/cragSummary.tsx
saferthanhouses: @vnugent Thanks for hte feedback! 
I've updated the PR with the button you suggested and removed the dynamic data fetching.

src/pages/crag/[id].tsx
```diff
@@ -84,101 +84,9 @@ export const getStaticProps: GetStaticProps<CragProps, { id: string }> = async (
...
-      updatedBy
-      createdAt
-      createdBy 
-    }
-  }`
 
-  const rs = await graphqlClient.query<{ area: AreaType }>({
-    query,
+  const rs = await graphqlClient.query<{ area: AreaType, getAreaHistory: ChangesetType[] }>({
+    query: QUERY_AREA_BY_ID,
```
Comment (vnugent) [inline src/pages/crag/[id].tsx:89]: Sorry about the extraordinarily belated response.

I think it's a great idea to use `QUERY_AREA_BY_ID` instead of inlining the long GQL query in the page.  However, the crag page is crashing due to undefined query fields. `QUERY_AREA_BY_ID` hasn't been updated lately to match the inline query.  All you need to do is updating QUERY_AREA_BY_ID with the deleted inline query.

Are you still available to fix this?  If not please let me know.  I'll update your PR by the end of this coming week.

Thank you!

vnugent (changes requested):

src/pages/crag/[id].tsx
Comment (saferthanhouses) [reply src/pages/crag/[id].tsx:89]: Hey @vnugent, sorry that I didn't catch that! I can update this pr today.

vnugent:
@all-contributors add @saferthanhouses for code & ideas

PR #774, opened 2023-04-03
Fix/ticklist revalidation

### LINKED ISSUE #727: Tick lists in opentacos don't reflect additions/deletions visible through API
I attempted to delete an ascent but was unable to do so.

## Expected Behavior

The tick should disappear from my tick list.

## Current Behavior

The tick remains in my tick list but I don't see the 
image
 button on the climb's page.

## Steps to Reproduce

1. Tick a climb.
2. Import ascents from mountain project.
3. See that tick is duplicate in log book.
4. Attempt to delete an ascent that is now a duplicate.
5. Check tick list again.  See that both ascents are still there.

## Context (Environment)

I added an ascent before trying to import from mountain project.  I am attempting to delete a duplicate tick.

musoke:
_Ready to go now_

I have added revalidation of tick lists after 10 seconds.  I can see in the log that the appropriate call to `getSaticProps` does happen once every 10 seconds when I continuously refresh the page.

However, this isn't yet sufficient to fix #727: the tick list doesn't change after adding ticks.  The data logged inside `getSaticProps` does not change either.

The data returned by querying the API directly through [hasura](https://cloud.hasura.io/public/graphiql?endpoint=https%3A%2F%2Fstg-api.openbeta.io%2F+&query=query+MyQuery+%7B%0A++userTicks%28userId%3A+%22c26b3c91-bb05-42a9-8f8c-b2209c55781b%22%29+%7B%0A++++_id%0A++++climbId%0A++++userId%0A++++name%0A++%7D%0A%7D%0A) does reflect ticks being added/removed.

Is there another layer of caching I'm missing?

### DIFF
--- a/src/js/graphql/api.ts
+++ b/src/js/graphql/api.ts
@@ -159,7 +159,8 @@ export const getTicksByUserAndClimb = async (climbId: string, userId: string): P
       variables: {
         climbId,
         userId
-      }
+      },
+      fetchPolicy: 'no-cache'
     })
 
     if (Array.isArray(res.data?.userTicksByClimbId)) {
@@ -177,7 +178,8 @@ export const getTicksByUser = async (userId: string): Promise<TickType[]> => {
       query: QUERY_TICKS_BY_USER,
       variables: {
         userId
-      }
+      },
+      fetchPolicy: 'no-cache'
     })
 
     if (Array.isArray(res.data?.userTicks)) {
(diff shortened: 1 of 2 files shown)


vnugent:
Can you try to add `fetchPolicy: 'no-cache'` to the 2 API calls?

https://github.com/OpenBeta/open-tacos/blob/develop/src/js/graphql/api.ts#L155-L190


Example of no-cache: https://github.com/OpenBeta/open-tacos/blob/develop/src/pages/climbs/%5Bid%5D.tsx#L426

musoke:
That works for me locally - thank you @vnugent !

PR #778, opened 2023-04-08
Added more LCOs to the static data file

CocoisBuggy:
Manual data entry may be a bit error prone in some scenarios. [The google sheet](https://docs.google.com/spreadsheets/d/1Mtb-q4gsg_fc-8B6ecDTlU8XBRjswvVs9HdbU2OTNUk/edit?usp=sharing) should help keep things in check for the time being and can be pulled down with a small script.

I did think about simply doing this at server runtime, but it seemed unnecessarily risky 

```py
import pandas as pd
from pydantic import BaseModel
import json

# Replace the below URL with the URL of your Google sheet
sheet_id = "1Mtb-q4gsg_fc-8B6ecDTlU8XBRjswvVs9HdbU2OTNUk"
lco_data = 0
lco_areas = 397743377
url = f"https://docs.google.com/spreadsheets/d/{sheet_id}/export?format=csv"
# Use pandas to read the sheet as a DataFrame
lco_sheet = pd.read_csv(url + f"&gid={lco_data}")
areas_sheet = pd.read_csv(url + f"&gid={lco_areas}")
lco_sheet = lco_sheet.where(pd.notnull(lco_sheet), None)
lcos = {}  # The actul index of the LCOs


class ClimbingCommunity(BaseModel):
    """
    This is the data model that we expect for each climbing community.
    """

    id: str
    areaIdList: list[str]
    name: str
    website: str
    facebook: str | None
    instagram: str | None
    report: str | None
    donation: str | None
    logo: str | None
    bio: str | None
    email: str | None


for idx, row in lco_sheet.iterrows():
    id = row["ID (Unique, assigned)"]

    lcos[id] = ClimbingCommunity(
        id=id,
        name=row["Name"],
        website=row["Website"],
        logo=row["Avatar / Logo"],
        instagram=row["Instagram"],
        facebook=row["Facebook"],
        donation=row["Donation"],
        report=row["Report"],
        bio=row["Bio"],
        email=row["Email"],
        areaIdList=[],
    )

# Now we can join the other sheet to this one
for idx, row in areas_sheet.iterrows():
    lcos[row["LCO ID"]].areaIdList.append(row["AreaID"])


with open("lco_data.json", "w") as f:
    f.write(json.dumps([dict(c) for c in lcos.values()], indent=4, sort_keys=True))
```


src/components/lco/PageBanner.tsx
```diff
@@ -8,14 +8,19 @@ export interface LCOProfileType {
   areaIdList: string[]
   /** Official name */
   name: string
-  /** IG Url */
-  instagram: string
   /** Official website */
   website: string
   /** Bad hardware report form url */
-  report?: string
+  report: string | null
```
vnugent, line 14: why couldn't we use `report?: string` which implicitly supports null, undefined or string?


if we use `report !== null` (equal signs) and report is `undefined` it will evaluates to true
 
```javascript
{
  report != null && // only 1 equal sign
            <a className='btn btn-primary btn-sm opacity-80' href={report} target='_blank' rel='noreferrer'>Report bad bolts</a>
}
```

vnugent (changes requested):
Thanks for this!   

Shouldn't we add an `enabled` flag to only show orgs that we have spoken to as discussed?

vnugent (changes requested):
I left some comment inline about using optional undefined as before.

PR #788, opened 2023-04-20
Manage organizations in Basecamp

zichongkao:
## Status
 - [x] Adding organization works.
 - [x] Update `addOrganization` GraphQL API so all data can be passed in rather than just `displayName` and `orgType`.
 - [x] Existing organizations can be shown in listview.
 - [x] Existing organizations can be updated from listview.
 - [ ] Implement pagination of organization listview.
 - [x] Listview updates when an org is added or updated.
 - [x] Validate `associatedAreaIds` and `excludedAreaIds` MUUID format.
 - [x] CSV download works.

Decided to punt on implementing pagination for the listview. We're likely to start off with a small number of organizations. I've tracked the task here: https://github.com/orgs/OpenBeta/projects/6?pane=issue&itemId=26799717

## Changes

Creates new tabs in Basecamp. Currently we only manage users, but this change adds organizations:
image

Hitting "+ Create" opens a modal where users can create new organizations:
image

"Update" uses the same modal, but with immutable data shown in the left side bar. Also observe MUUID validation at work:
image

On mobile:
image


src/components/ui/form/Select.tsx
```diff
@@ -0,0 +1,105 @@
+import { Controller, RegisterOptions, useFormContext, UseFormReturn } from 'react-hook-form'
```
zichongkao: Currently we only have `Input`, `Checkbox` and a few others. `Select` would complete the lineup. But the standard `<select><option></option></select>` elements look very out of place amidst the rest of our elements, so I co-opted our `DropdownMenu` component and made it compatible with react-hook-form using `Controller`.

src/components/basecamp/CreateUpdateModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+interface CreateUpdateModalProps {
+  isOpen: boolean
+  onClose?: () => void
+  contentContainer: ReactElement
+}
+
+/**
+ * Modal for updating and creating objects (areas, organizations, etc) in our datastores.
+ */
+export default function CreateUpdateModal ({
```
vnugent, line 13: We're deprecating `@headlessui/react` and moving to `radix-ui`

Can you wrap the edit screen with a MobileDialog?  You can see an example here 

https://github.com/OpenBeta/open-tacos/blob/develop/src/components/edit/Triggers.tsx#L95

---

src/components/ui/form/Select.tsx
vnugent: we're using DaisyUI for styling: https://daisyui.com/components/select/

vnugent (changes requested):

src/components/basecamp/Organizations.tsx
```diff
@@ -0,0 +1,140 @@
...
+      {isAuthorized && <OrganizationTable />}
+    </>
+  )
+}
+
+const OrganizationTable = (): JSX.Element => {
+  const [modalOpen, setModalOpen] = useState(false)
+  const [focussedOrg, setfocussedOrg] = useState<OrganizationType | null>(null)
+
+  const { data, error } = useQuery(
```
zichongkao, line 38: The User page uses `useSWR` but I don't think it works here because that's calling the Auth0 API, while here our Apollo GraphQL client already has caching and refetching logic. 

I wasn't sure whether to fetch data server-side. But I figure client-side is fine given that users are admins, so performance is less important. Also data access is more adhoc so less opportunities to benefit from page-level caching. 

https://nextjs.org/docs/basic-features/data-fetching/client-side

src/components/basecamp/Users.tsx
```diff
@@ -36,7 +36,10 @@ export default function Users (): JSX.Element {
 
 const LinkProfile = ({ nick }: {nick: string}): JSX.Element => <Link href={`/u/${nick}`}><a className='link-primary'>{nick}</a></Link>
 
-const fetcher = async (url: string): Promise<any> => (await axios.get(url)).data
+const fetcher = async (url: string): Promise<any> => {
+  console.log('fetcher', url)
```
zichongkao: A bunch of logs I still need to remove, but ok for first pass at review!

src/components/ui/form/Select.tsx
zichongkao: I updated the classNames to use "select" classes instead of "btn" ones. Hope that was what your suggestion meant!

zichongkao:
Thinking about this more, I feel like in the long run, Basecamp could be a standalone repo and an app of it's own. That would allow it to become super powerful without being constrained by the public-facing product: Different authentication process, no need mobile support (admins can edit on desktop), etc.

This would enable us to engineer the underlying data to the highest possible standards. We could develop tools to visualize and rework area hierarchies, take bulk actions on db objects, highlight and respond to contentious edits, etc.

vnugent:
>  Implement pagination of organization listview.

I'd not worry about implementing pagination right now since there are so few LCOs.  The public facing edit log https://openbeta.io/edit page could use some pagination.

src/components/basecamp/OrganizationForm.tsx
```diff
@@ -0,0 +1,263 @@
...
+    message: 'Minimum 2 characters.'
+  }
+}
+
+/**
+ * Detects if string is in uuid-mongodb's "relaxed" hex format.
+ * @param s input string
+ * @returns
+ */
+const isMuuidHexStr = (s: string): boolean => {
```
vnugent: the uuid library (already in used) has a uuid v4 validation:  https://www.npmjs.com/package/uuid#uuidvalidatestr

src/components/basecamp/OrganizationForm.tsx
```diff
@@ -0,0 +1,263 @@
...
+  /**
+   * Routes to addOrganization or updateOrganization GraphQL calls
+   * based on whether there was an existing org to update or not.
+   * @param formProps Data populated by the user in the form
+   * @returns
+   */
+  const submitHandler = async (formProps: HtmlFormProps): Promise<void> => {
+    if (existingOrg === null) {
+      if (formProps.displayName == null) {
+        console.error('`displayName` cannot be null when creating organization.')
```
vnugent: since you already specify a validation rule for displayName, react-hook-form won't call `submitHandler()` until the user has fixed all errors.

src/components/basecamp/OrganizationForm.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const { handleSubmit, formState: { isSubmitting } } = form
+
+  /**
+   * Routes to addOrganization or updateOrganization GraphQL calls
+   * based on whether there was an existing org to update or not.
+   * @param formProps Data populated by the user in the form
+   * @returns
+   */
+  const submitHandler = async (formProps: HtmlFormProps): Promise<void> => {
+    if (existingOrg === null) {
```
vnugent: I'd use 2 equal signs, to catch both nullness and undefined check.  If react-hook-form returns `undefined` then existingOrg is not null.

`if (existingOrg == null) {`

src/components/basecamp/OrganizationForm.tsx
```diff
@@ -0,0 +1,263 @@
...
+      <div className='basis-2/3 grow'>
+        <FormProvider {...form}>
+          <form onSubmit={handleSubmit(submitHandler)} className='min-w-[16em]'>
+            <Input
+              label='Display Name:'
+              name='displayName'
+              placeholder='Climbing Org'
+              registerOptions={DISPLAY_NAME_FORM_VALIDATION_RULES}
+            />
+            {existingOrg === null &&
```
vnugent: same here, please use == (2 equal signs)

src/components/basecamp/OrganizationForm.tsx
```diff
@@ -0,0 +1,263 @@
...
+    if (existingOrg === null) {
+      if (formProps.displayName == null) {
+        console.error('`displayName` cannot be null when creating organization.')
+        return
+      }
+      if (formProps.orgType == null) {
+        console.error('`orgType` cannot be null when creating organization.')
+        return
+      }
+      const input = removeNullUndefined({
```
vnugent: if you want to submit only dirty fields you can use `dirtyFields.fieldName === true`

Climb edit form pre-submit:
https://github.com/OpenBeta/open-tacos/blob/develop/src/pages/climbs/%5Bid%5D.tsx#L180-L187

musoke (approved):
I'm entirely unqualified to review this, but the console.logs are all the stood out to me.
I'm happy if Viet is happy.

zichongkao:
@vnugent Thanks for the feedback! Using `dirtyFields`, and `uuidValidation` cleaned up the code really nicely!

src/components/basecamp/Organizations.tsx
vnugent: For admin dashboard client side fetching is perfectly fine :)

Issue #750, opened 2023-03-23
How to encourage people to set their user id and display name?


vnugent:
**What would you like to be able to do? Please describe.**
When looking at the user profile edit page, most people don't know:

1.  You can change the auto-generated user id to your own unique id (no space allowed, similar to IG or Twitter handle)
2. Display name can be a human-friendly name with space Ex:  "Jane Doe".  A lot of people would put "jane.doe".

**How important is this to you** (Please pick one)
- Nice-to-have

### To do:
- [ ] Suggest small UI changes to encourage people to set a user name and a display name

<img width="650" alt="Screenshot 2023-03-23 at 11 57 35 AM" src="https://user-images.githubusercontent.com/3805254/227184306-f8bb6b3f-a833-4952-be2e-01209feff26c.png">

clintonlunn:
@vnugent is there a way to figure out if the username/display name is the default randomly generated name? It might be nice to only tell the user that they can change their name if they haven't already changed it before.

vnugent:
The auto-generated name has the following format: `<random adjective>_ <random color>_<random animal>`.    

What we can also do is somehow detect if it's the first login, the code would clear the auto-generated name, forcing the user to change it if they want to update the profile.

clintonlunn:
@vnugent 

Ok there is a [loginsCount](https://github.com/OpenBeta/open-tacos/blob/e6bdcc17f97a2e155c5dd34cc5e8185da2efd743/src/js/types/User.ts#L4) property that comes in with the [userProfile](https://github.com/OpenBeta/open-tacos/blob/f043f174e76a5d632a2ff4778955c59665fd544c/src/components/users/PublicProfile.tsx#L24) property. 

How about we open a tooltip on the user's first login?
![image](https://user-images.githubusercontent.com/24685932/233225116-00c12411-c299-4f5b-a543-b8790fcb8ea0.png)

I can take issue.

clintonlunn:
@vnugent how would you feel about routing the user directly to the edit profile page after creating a profile? Then, if it's their first login, we can activate some sort of a stepper component

![image](https://user-images.githubusercontent.com/24685932/234719356-4e4a2016-596f-459f-987a-7f81381e5163.png)


and if it's not their first login, they'll only see the normal form
![image](https://user-images.githubusercontent.com/24685932/234719697-ade1cf37-aaae-4c6b-ab3a-83bbae716033.png)

vnugent:
New users are now required to pick a username.

Issue #746, opened 2023-03-22
Email validation in user signup flow


actuallyyun:
**What would you like to be able to do? Please describe.**

After signing up, I was prompted to check my email to validate my email address. After clicking on the link in my email, I was directed to this url `https://openbeta.us.auth0.com/u/email-verification?ticket=dGXPhgEpgEcGbJ1GMgjGc5zGdiCHBlrl#`. A success message was shown which was great. 

It would be nice to be redirected to sign in page or the main page after this though. 
![Screenshot 2023-03-22 at 15 19 03](https://user-images.githubusercontent.com/87448230/226933328-31563f29-d583-4b53-b588-be2af613917a.png)

**How important is this to you** (Please pick one)
Nice-to-have

vnugent:
@all-contributors add @actuallyyun for ideas

actuallyyun:
Hi @vnugent  I would like to work on this as my first issue if this sounds good to you.

vnugent:
> Hi @vnugent I would like to work on this as my first issue if this sounds good to you.

Definitely.  One thing though, we use Auth0.com as our identity/authentication provider.   I'm not sure if we have control over that dialog unless we set up a custom email provider in Auth0 dashboard (we're currently using their free email service).  

https://auth0.com/docs/customize/email

Send me an email (viet at openbeta dot io) and I'll reply with additional API keys and give you access to the Auth0 test instance.  

IMO first PR should be fun.  In my experience Auth0 docs can be very confusing and sometimes it requires getting help on [their forum](https://community.auth0.com/). Do you want to do a bit of investigation to see if it's possible to redirect the user?

actuallyyun:
yeah good point, I will look into their docs and see if it's possible.

actuallyyun:
This seems to be one solution to this: https://community.auth0.com/t/need-to-update-the-redirect-url-for-password-reset/9626  It doesn't require code change, but configurations in `AuthO` settings.

actuallyyun:
@vnugent If you could share the `Auth0` account, I could go in and try to fix this.

vnugent:
> @vnugent If you could share the `Auth0` account, I could go in and try to fix this.

email me (viet at openbeta dot io) and I'll reply with additional API keys, etc

actuallyyun:
Hey @vnugent After some digging in their docs, I think one solution is to config a default login route, then their universal login flow will add this to the email validation result, and the user can click and get redirected to the login page.
[...]

PR #789, opened 2023-04-22
Encourage first time users to change their username #750

clintonlunn:
I added a check to see if it was the user's first login, and if it is, there will be a tooltip to prompt the user to change their username away from the autogenerated one.

I also added unit tests to EditProfileButton and forOwnerOnly to ensure those are working as expected. 

image

### DIFF
--- a/src/components/users/EditProfileButton.tsx
+++ b/src/components/users/EditProfileButton.tsx
@@ -16,7 +16,7 @@ interface EditProfileButtonProps extends WithOwnerProfile {
 function EditProfileButton ({ loginsCount }: EditProfileButtonProps): JSX.Element {
   return (
     <>
-      {loginsCount === 2 // the loginsCount is 2 after the user creates their account
+      {loginsCount === 1
         ? (
           <Tooltip
             side='bottom'
@@ -32,7 +32,7 @@ function EditProfileButton ({ loginsCount }: EditProfileButtonProps): JSX.Elemen
             />
           </Tooltip>
           )
-        : loginsCount !== undefined && loginsCount > 0
+        : loginsCount !== undefined && loginsCount > 1
           ? (
             <Button
               href='/account/edit'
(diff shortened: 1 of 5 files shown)


clintonlunn:
I am struggling with testing this because of the how the signup functionality routes you to stg.openbeta.io instead of keeping you in the preview url's, which might run up the loginCount to 1 before I can test on the preview build. I believe the loginsCount after a user creates a profile would be either 0 or 1, but a bit more confident on 1.

src/components/users/EditProfileButton.tsx
```diff
@@ -1,14 +1,48 @@
...
+/**
+ * EditProfileButton component displays a "Edit" button with optional tooltip based on loginsCount prop.
+ * @param {EditProfileButtonProps} loginsCount - The number of times a user has logged in since account creation.
+ * @returns {JSX.Element} - JSX element representing the EditProfileButton component.
+ */
+function EditProfileButton ({ loginsCount }: EditProfileButtonProps): JSX.Element {
   return (
-    <Button href='/account/edit' label='Edit' variant={ButtonVariant.OUTLINED_DEFAULT} size='sm' />
+    <>
+      {loginsCount === 1
```
vnugent, line 19: what if the user ignores the tooltip and closes the browser?  The next time will they see the reminder?

vnugent (commented):
what do you think if we create a proper onboarding process? 
- create a new screen to prompt the user to do x, y, z after signing up
- create a new attribute in the user profile's metadata to track onboarding completion. This way if they chose to ignore it, they will be prompted again next time. 

FYI -we have the ability to direct newly signed up users to a url
https://github.com/OpenBeta/open-tacos/issues/746

src/components/users/EditProfileButton.tsx
clintonlunn: You're right, I didn't really consider that. I think some sort of formalized onboarding process would be beneficial

PR #795, opened 2023-04-23
adding unit test to forOwnerOnly #794

### LINKED ISSUE #794: Add unit tests for forOwnerOnly higher order function
Currently, there are no unit tests for the [forOwnerOnly](https://github.com/OpenBeta/open-tacos/blob/880f942aa80a3aaf66ecb51363856fe5473b2c96/src/js/auth/forOwnerOnly.ts) higher order function, which is used to restrict access to certain components or functions to a specific owner. The function takes in a component as an argument and returns a new component that only renders if the current user is the owner.

This test would be especially useful because of the need to mock the response from the usePermission hook. I think that anywhere else that needs to use a component with this forOwnerOnly() function would also need to mock the userPermission() response to get expected behavior. 

**Importance**: Nice-to-have

clintonlunn:
This pull request adds unit tests for the forOwnerOnly higher order function, which is used to restrict access to certain components or functions to a specific owner. The tests verify that the function behaves as expected and that it correctly restricts access to components based on if isAuthorized returns `true` from the mocked usePermission hook.

### DIFF
--- a/src/js/__tests__/auth/forOwnerOnly.tsx
+++ b/src/js/__tests__/auth/forOwnerOnly.tsx
@@ -1,0 +1,64 @@
+import React from 'react'
+import '@testing-library/jest-dom/extend-expect'
+
+import { render } from '@testing-library/react'
+import forOwnerOnly from '../../auth/forOwnerOnly'
+import usePermissions from '../../hooks/auth/usePermissions'
+import { IUserProfile, WithOwnerProfile } from '../../types/User'
+
+// Mock the usePermissions hook
+jest.mock('../../hooks/auth/usePermissions')
+
+const dummyOwnerProfile: IUserProfile = {
+  uuid: '12345678-1234-1234-1234-123456789012',
+  roles: ['user'],
+  loginsCount: 1,
+  name: 'John Doe',
+  nick: 'johndoe',
+  bio: 'Climber and outdoor enthusiast',
+  website: 'https://johndoe.com',
+  ticksImported: true,
+  collections: {
+    climbCollections: {
+      'John\'s Climbs': ['climb1', 'climb2', 'climb3']
+    },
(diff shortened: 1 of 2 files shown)


src/js/__tests__/auth/forOwnerOnly.tsx
```diff
@@ -0,0 +1,64 @@
...
+    areaCollections: {
+      'John\'s Areas': ['area1', 'area2', 'area3']
+    }
+  },
+  email: 'john@example.com',
+  avatar: 'https://example.com/john/avatar.jpg',
+  authProviderId: 'auth0|123456789012345678901234'
+}
+// Create a mock component to be used with forOwnerOnly
+const MockComponent: React.FC<WithOwnerProfile> = () => <div>Mock Component</div>
```
MichaelDimmitt, line 34: consider:
`const MockComponent = (unusedProps: WithOwnerProfile) => <div>Mock Component</div>`

src/js/__tests__/auth/forOwnerOnly.tsx
```diff
@@ -0,0 +1,64 @@
...
+// Create a component wrapped with forOwnerOnly
+const WrappedComponent = forOwnerOnly(MockComponent)
+
+describe('forOwnerOnly', () => {
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('should render the component if isAuthorized is true', async () => {
+    (usePermissions as jest.Mock).mockReturnValue({ isAuthorized: true })
```
MichaelDimmitt, line 45: // jest.Mock(usePermission) 
not sure the actual constructor here so pseudocode above. the pseudocode would in theory should have the correct prototype, previously you are just adding the type which does nothing. The only reason it is working is because of some jest global environment magic.

PR #802, opened 2023-04-27
set up a new user profile edit workflow #750

clintonlunn:
An attempt at a new user process when it's their first login. 

However, this workflow will persist for as long as their login is 1. I'm not sure how to address that so far. Cookies, sessionstorage (but would trigger again on first login but separate session)

image


vnugent:
@clintonlunn 

I have a suggestion:
- Create a reusable onboarding progress component (there's a simple one under the user profile - see screencap below)
- The progress can be a checklist or step progress with links/buttons for the user to take action
- Wrap the progress component in a `Toast` which we continue to show until the login count > 2 or 3?  Users can always dismiss the toast.  In the future we can introduce a check to make sure the user has changed their user name.

Check list
- [ ] Change user name
- [ ] Upload at least 3 photos

---
Screenshot 2023-05-03 at 9 30 39 AM

clintonlunn:
@vnugent I think I am getting closer. I removed the stepper and created an onboarding checklist that pops up as a toast notification on the first few user logins. The username checklist item doesn't really make sense, but i'm still hoping to find a way to figure out if the user has a default username. 

image

clintonlunn:
@vnugent sorry i had to step away from this issue for a little while. Looks like you've got  a pr that you're working on that would cover this issue. Should I close this pr?

vnugent:
Yes, we can close this one.   The new flow will block newly registered users from going to the site until they set a username.
-  Backend https://github.com/OpenBeta/openbeta-graphql/pull/290
 - Frontend https://github.com/OpenBeta/open-tacos/pull/838

PR #782, opened 2023-04-17
issue #763 LCO List Component based on new Figma design

actuallyyun:
#### Empty State 
- web
Screenshot 2023-04-19 at 11 47 35

- mobile
Screenshot 2023-04-19 at 11 47 52

#### Happy Path
- web
Screenshot 2023-04-19 at 11 53 49

- mobile
Screenshot 2023-04-19 at 11 54 08

-Tooltip

Screenshot 2023-04-24 at 10 46 59

PR #805, opened 2023-04-29
Fix getChangeHistory

zichongkao:
In response to breakage reported here: https://github.com/OpenBeta/open-tacos/pull/782.

In working on this, I considered whether we should be showing Organizations history in the general changelog. I feel it could go either way:
-  If we think of the changelog as helping folks keep tabs on user-executed changes so that we can fix them if there is abuse, then Organization changes shouldn't be there because these are done by LCO admins whom we trust more and are less at risk of abuse.
- If we think of the changelog as showing folks what's going on the site that might be of public interest, then since organizations are highly visible, it makes sense to show changes to them.

In the end, I figured we can just expand functionality to show a change (including those for organizations) if it exists. If we want to remove organization changes, the best place is probably at the GraphQL query layer, by adding a filter. 

The current filter doesn't support `kind`. The problem is that a changelog of a certain kind can involve documents of multiple kinds. Eg. add climb (changelog `kind` == 'climbs') involves inserting a climb (document `kind` == 'climbs') and updating the parent area (document `kind` = 'areas'). It's not clear whether we should include the parent area document change when filtering for climbing changelogs. So there's some nuance here to think about.

For context, `kind` is stabilized on the backend in this https://github.com/OpenBeta/openbeta-graphql/pull/277.

### DIFF
--- a/src/components/edit/RecentChangeHistory.tsx
+++ b/src/components/edit/RecentChangeHistory.tsx
@@ -3,7 +3,7 @@ import Link from 'next/link'
 import { PlusIcon, UserCircleIcon, MinusIcon, PencilIcon, PencilSquareIcon, MinusCircleIcon, PlusCircleIcon } from '@heroicons/react/24/outline'
 import { formatDistanceToNow } from 'date-fns'
 
-import { ChangesetType, ChangeType, AreaType, ClimbType } from '../../js/types'
+import { ChangesetType, ChangeType, AreaType, ClimbType, OrganizationType, DocumentTypeName } from '../../js/types'
 
 export interface RecentChangeHistoryProps {
   history: ChangesetType[]
@@ -43,6 +43,7 @@ const ChangesetRow = ({ changeset }: ChangsetRowProps): JSX.Element => {
             <React.Fragment key={change.changeId}>
               <AreaChange {...change} />
               <ClimbChange {...change} />
+              <OrganizationChange {...change} />
             </React.Fragment>))}
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


src/components/edit/RecentChangeHistory.tsx
```diff
@@ -43,15 +43,16 @@ const ChangesetRow = ({ changeset }: ChangsetRowProps): JSX.Element => {
...
         </div>
       </div>
     </div>
   )
 }
 
-const ClimbChange = ({ changeId, fullDocument, updateDescription, dbOp }: ChangeType): JSX.Element | null => {
-  if ((fullDocument as ClimbType)?.name == null) {
+const ClimbChange = ({ changeId, fullDocument, updateDescription, dbOp, kind }: ChangeType): JSX.Element | null => {
+  if (kind !== DocumentKind.climbs) {
```
zichongkao: More robust to use `__typename` than unique document fields: https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces/#querying-a-union

src/components/edit/RecentChangeHistory.tsx
```diff
@@ -51,7 +52,7 @@ const ChangesetRow = ({ changeset }: ChangsetRowProps): JSX.Element => {
 }
 
 const ClimbChange = ({ changeId, fullDocument, updateDescription, dbOp }: ChangeType): JSX.Element | null => {
-  if ((fullDocument as ClimbType)?.name == null) {
+  if (fullDocument.__typeName !== DocumentTypeName.Climb) {
```
zichongkao, line 55: More robust to use `__typename` than unique document fields: https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces/#querying-a-union

zichongkao:
Note that builds will keep failing until newest openbeta-graphql changes are deployed to staging. Otherwise during build, we prepare static props which requires calling the staging API and that fails because the old code doesn't expect Organization document changes.

PR #806, opened 2023-05-01
Update users from Basecamp page

zichongkao:
Aim of this PR is to allow admins to go to Basecamp, designate some users as LCO admins and tag them with the LCOs they are admins of.

I redesigned `/basecamp/users` as well.
- Use `<table>` instead of `<div className='grid>`. Rationale is that grids lack row/column structure -- they work well for dynamically arranged cards and such. But in this case, I wanted to do row-level formatting which is v messy with grids. Let me know if I missed something, and if grids are actually the way to go here!
- I added email address search functionality so that we can find the accounts we want to grant LCO admin status to. Otherwise it's just one big list and we would have to page through it to find the desired ones. In theory we could search according to other identifiers, but I'm guessing most admins will have their organization's email domains which will make searching easy.

Before:
image

After:
image

User editor modal:
image

Form blocks adding organization if user is not `org_admin`:
image

Also updated the Organization page to match the updated design:
image

## Deployment
NOTE: Before deploying this to prod, we need to update Auth0 so that the prod client has the newly request scope: `read: roles`. Failure to do this will break prod because the Auth0ManagementClient will start failing from requesting a scope that was not granted to it.

### DIFF
--- a/src/components/basecamp/UserForm.tsx
+++ b/src/components/basecamp/UserForm.tsx
@@ -1,0 +1,135 @@
+import type { User } from 'auth0'
+import clx from 'classnames'
+import { TextArea } from '../ui/form'
+import { useForm, FormProvider } from 'react-hook-form'
+import { MUUID_VALIDATION, conjoinedStringToArray } from './utils'
+import axios from 'axios'
+import useSWR from 'swr'
+import { Role } from 'auth0'
+import { UserRole } from '../../js/types'
+import MultiSelect from '../ui/form/MultiSelect'
+import { useEffect } from 'react'
+
+interface UserFormProps {
+  user: User
+  onClose: () => void
+}
+
+interface HtmlFormProps {
+  roles: string[]
+  conjoinedOrgAdminOrgIds: string // Form will return one large conjoined string
+}
+
+const fetcher = async (url: string): Promise<any> => (await axios.get(url)).data
+
+/*
+ * Form for updating users.
+*/
+export default function UserForm ({ user, onClose }: UserFormProps): JSX.Element {
+  const userId = user.user_id
+  if (userId == null) {
+    return <div>Can't load user. Missing user_id.</div>
+  }
+  const { data: roles, error, mutate } = useSWR<Role[]>(`/api/basecamp/userRoles?userId=${userId}`, fetcher)
+  if (error != null) {
+    return <div>{error}</div>
+  }
+  const defaultValues = {
+    roles: roles?.map(r => r.name),
+    conjoinedOrgAdminOrgIds: user.user_metadata?.orgAdminOrgIds?.join(', ') ?? ''
+  }
+
+  // React-hook-form declaration
+  const form = useForm<HtmlFormProps>({
+    mode: 'onBlur',
+    defaultValues
+  })
+  const { handleSubmit, reset, watch, formState: { isSubmitting, dirtyFields } } = form
+
+  useEffect(() => {
+    /* Without this, even after useSWR's refetch produces new data, defaultValues.roles remains
+     * undefined and falls back to the component's default (which for roles is []).
+     */
+    if (roles != null) {
+      reset(
+        defaultValues,
+        { keepErrors: true, keepDirty: true }
+      )
+    }
+  }, [roles, user.user_metadata?.orgAdminOrgIds]) // Dependencies of defaultValues
+
+  /**
+   * Call various Auth0 APIs to update new user data.
+   * @param formProps Data populated by the user in the form
+   * @returns
+   */
+  const submitHandler = async ({
+    conjoinedOrgAdminOrgIds,
+    roles
+  }: HtmlFormProps): Promise<void> => {
+    if ((dirtyFields?.roles as boolean | undefined) === true) {
+      await axios.post(`/api/basecamp/userRoles?userId=${userId}${roles.map(r => `&roles=${r}`).join('')}`)
+    }
+    if (dirtyFields?.conjoinedOrgAdminOrgIds === true) {
+      const orgAdminOrgIds = conjoinedStringToArray(conjoinedOrgAdminOrgIds)
+
+      // Ideally we'd also validate that these orgIds match actual organizations.
+      await axios.post(`/api/basecamp/user?userId=${userId}${orgAdminOrgIds.map(o => `&orgAdminOrgIds=${o}`).join('')}`)
+    }
+    void mutate() // Have SWR refetch data. Auth0 role endpoints don't return the latest state, so we can't set manually.
+    onClose()
+  }
+  return (
+
+    <div className='px-8 pt-12 pb-4 flex flex-row flex-wrap'>
+      <div className='basis-1/3 shink-0 grow min-w-[10em] pr-4'>
+        <h2>User Editor</h2>
+        <div className='mt-4'>
+          <div className='mt-4 break-words'>
+            <h4>userId</h4>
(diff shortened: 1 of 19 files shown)


vnugent:
@zichongkao can you please redact emails?

zichongkao:
@vnugent Just to check -- you mean to redact the screenshots right? I've just done that. Let me know if you mean implement email redaction in Basecamp.

vnugent:
> t Just to check -- you mean to redact the screenshots right? I've just done that. Let me know if you mean implement email redaction in Basecamp.

@zichongkao yes, just the screenshot when sharing. thanks!

src/components/basecamp/UserForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+    mode: 'onBlur',
+    defaultValues: {
+      roles: roles?.map(r => r.name),
+      conjoinedOrgAdminOrgIds: user.user_metadata?.orgAdminOrgIds?.join(', ') ?? ''
+    }
+  })
+  const { handleSubmit, setValue, watch, formState: { isSubmitting, dirtyFields } } = form
+
+  useEffect(() => {
+    if (roles != null) setValue('roles', roles.map(r => r.name as string))
```
zichongkao: Seems like the `useForm` hook sets the defaultValues on first render, and then never revisits it. However, useSWR serves up stale data on first render and refetches in the background. When the refetch returns, useForm no longer updates the values of the form leaving the roles with inaccurate data. This effect fixes that by explicitly setting the value.

src/components/ui/ComboBox.tsx
```diff
@@ -6,29 +6,43 @@ function classNames (...classes): string {
...
-export default function ComboBox ({ options, value, onChange, label }: ComboBoxProps): JSX.Element {
+interface ComboBoxMultiProps {
+  options: ValueObj[]
+  value: ValueObj[]
+  onChange: (value: ValueObj[]) => void
+  label: string
+  multiple: true
+  selectClassName?: string
+}
+type ComboBoxProps = ComboBoxSingleProps | ComboBoxMultiProps
```
zichongkao, line 30: A bit of a hack to reuse the `ComboBox`. But perhaps I have gone too far and should have create a brand new component for multi-selections. Lmk if you think so and I can make the change.

src/components/ui/ComboBox.tsx
```diff
@@ -6,29 +6,43 @@ function classNames (...classes): string {
...
+
+export default function ComboBox (props: ComboBoxProps): JSX.Element {
+  const { options, value, onChange, label, selectClassName = '' } = props
   return (
-    <Listbox value={value} onChange={onChange}>
+    <Listbox value={value} onChange={onChange} multiple={'multiple' in props ? props.multiple : false}>
       {({ open }) => (
         <>
-          <Listbox.Label className='block text-sm font-medium text-gray-700 mt-2'>{label}</Listbox.Label>
+          {label !== '' && <Listbox.Label className='block text-sm font-medium text-gray-700 mt-2'>{label}</Listbox.Label>}
```
zichongkao, line 38: So that `MultiSelect` can insert its own label with formatting consistent across all form inputs.

src/js/auth/ManagementClient.ts
```diff
@@ -11,25 +11,31 @@ export const auth0ManagementClient = new Auth0MgmtClient({
   domain: issuer.replace('https://', ''),
   clientId: mgmtClientId,
   clientSecret: mgmtClientSecret,
-  scope: 'read:users update:users create:users'
+  scope: 'read:users update:users create:users read:roles'
```
zichongkao, line 14: See my note in the PR write-up. Very important we change the client scope during deployment.

src/js/auth/ManagementClient.ts
```diff
@@ -116,3 +122,49 @@ export const extractUpdatableMetadataFromProfile = ({ name, nick, bio, website,
...
+export const getUserRoles = async (userId: string): Promise<Role[]> => {
+  return await auth0ManagementClient.getUserRoles({ id: userId, page: 0, per_page: 50 })
+}
+
+/**
+ * Sets roles for the user, making multiple calls to Auth0 since there is no consolidated set call.
+ * @param userId Auth0 internal user id. Ex: auth0|234879238023482995
+ * @param roles Array of role names (Ex: 'editor'), not Auth0 role IDs (Ex: 'rol_ds239fjdsfsd')
+ */
+export const setUserRoles = async (userId: string, roles: string[]): Promise<void> => {
```
zichongkao, line 139: Kinda unfortunate that we have to make three API calls to set roles. Challenge is that Auth0 only has `assignRolesToUser` which strictly add roles and `removeRolesFromUser` which strictly removes roles. So we need at least two API calls.

The `getRoles` call, I'm using to pull role IDs. In theory we could hardcode this as a set of constants, but the risk is that they get out of sync. 

I could also do a `Promise.all` for the assign and remove calls to try to run them concurrently. Lmk if you'd like that.

src/js/types/User.ts
```diff
@@ -28,7 +38,18 @@ export interface IWritableUserMetadata {
...
+ * `user_metadata` as seen in the Auth0 datastore,
+ * before we add fields during post-login actions.
+ */
+export type IUserMetadataOriginal = IWritableUserMetadata & IReadOnlyUserMetadata
+
+/* `roles` are added by a post-login action and are not in
+ * the user's `user_metadata` object in Auth0.
+ */
+export type IUserMetadata = IUserMetadataOriginal & {
+  roles: UserRole[]
```
zichongkao, line 51: Pull this out because this field cannot be set on the user since it's not actually in the user's `user_metadata` field.

zichongkao:
I wrote zero tests -- lmk if you think I should add some. Could add some for the updated `ComboBox` and `Input`. Potentially also `MultiSelect`.

src/components/basecamp/UserForm.tsx
vnugent: You can call react-hook-form `reset()` and passing in new values to the form. 
- https://stackoverflow.com/questions/62242657/how-to-change-react-hook-form-defaultvalue-with-useeffect
- https://react-hook-form.com/api/useform/reset/

src/js/auth/ManagementClient.ts
vnugent: I think it's fine making 2 separate awaits

src/js/auth/ManagementClient.ts
vnugent: can you post a screenshot of Auth0 test tenant where you added the role? (Auth0 dashboard is so complex. I want to make sure I'm making the same change in prod).

src/components/ui/ComboBox.tsx
vnugent: I think we're good for now.  If there's a need for multi-select combobox again we can create a reusable one.

vnugent (changes requested):
I'd make this one change, using `reset()` to refresh form and we're good to merge.  Thanks!
https://github.com/OpenBeta/open-tacos/pull/806/files/0603a035acc2e15895eb11a440cf45523d402ced#r1183315679

src/js/auth/ManagementClient.ts
zichongkao: image

zichongkao:
@vnugent This one's ready for your review!

vnugent:
can you please rebase? thanks.  we're good to merge.

zichongkao:
@vnugent rebased!

PR #813, opened 2023-05-07
feat: Lco profile card popup

actuallyyun:
desktop:
Screenshot 2023-05-18 at 14 19 14
Screenshot 2023-05-18 at 14 19 21

mobile:
Screenshot 2023-05-18 at 14 19 39
Screenshot 2023-05-18 at 14 19 47

lco list:
Screenshot 2023-05-18 at 14 18 31

@vnugent I added one change to `MobileDialog` to hide the close dialog button. This affects all dialogs. Based on the discussion I had with AJ, he thinks that on bigger screens, where there's space outside of the dialog, user could just click outside of the dialog to close it. I think it makes sense. If you don't think it's a good idea, I could reverse this change.

### DIFF
--- a/src/components/lco/PageBanner.tsx
+++ b/src/components/lco/PageBanner.tsx
@@ -12,7 +12,7 @@ export interface LCOProfileType {
   areaIdList: string[]
   /** Official name */
   name: string
-  description: string
+  description?: string
   email?: string
   facebook?: string
   /** IG Url */
@@ -33,7 +33,7 @@ const getLcoList = (orgs): LCOProfileType[] => {
       id: org.orgId,
       areaIdList: org.associatedAreaIds ?? [],
       name: org.displayName,
-      description: org.content?.description ?? '',
+      description: org.content?.description,
       email: org.content?.email,
       website: org.content?.website,
       facebook: org.content?.facebookLink,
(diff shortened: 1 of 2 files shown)


src/components/lco/PageBanner.tsx
```diff
@@ -124,6 +136,100 @@ interface ContentProps {
...
+                target='_blank'
+                rel='noreferrer'
+              >
+                Make a Donation <ArrowUpRightIcon className='ml-2 w-4 h-4' />
+              </a>
+            )}
+          </div>
+          <div className='card-actions '>
+            <a
+              className='btn btn-primary opacity-80 btn-solid bg-red-500 border-0'
```
vnugent: For the primary CTA should we just use `btn btn-primary btn-solid` (dark black) instead of introducing a new custom color?

actuallyyun: Yeah sure! I'll change it.

actuallyyun: Screenshot 2023-05-16 at 17 21 47
Screenshot 2023-05-16 at 17 22 00

src/components/lco/PageBanner.tsx
```diff
@@ -92,56 +104,100 @@ const IndividualBanner: React.FC<ContentProps> = ({ profile }) => (
...
+        <a
+          className='underline pb-5'
+          href={website}
+          target='_blank'
+          rel='noreferrer'
+        >
+          <p className='text-sm'>{website}</p>
+        </a>
+        <p>
+          {description}
```
actuallyyun: Currently, description data does not have paragraph breaks. I could at the fe parse it to proper form, or, do you think it's more appropriate to clean it in the be? @zichongkao

src/components/lco/PageBanner.tsx
```diff
@@ -92,56 +104,100 @@ const IndividualBanner: React.FC<ContentProps> = ({ profile }) => (
...
-//      <div className='card-actions bg-base-100 flex justify-end p-4'>
-//        <a className='btn btn-primary btn-sm opacity-80' href={report} target='_blank' rel='noreferrer'>Report bad bolts</a>
-//      </div>
-//    </div>
-//  )
-// }
+      <div className='gap-10 md:col-span-5'>
+        <p className='text-base-content/60 font-semibold'>Local Climbing Organization</p>
+        <h2 className='card-title my-2 uppercase'>{name}</h2>
+        <a
```
zichongkao: As you noted during the call, this can be null too, so it'd be good to check this and hide if null.

actuallyyun: You mean lco name? I thought name cannot be null. I addded check against other fields.

zichongkao: you're good -- i meant website!

zichongkao (approved):
Looks good :D Thanks for updating this!

PR #820, opened 2023-05-10
Added Toast to Update User Profile Form

### LINKED ISSUE #814: Replace Snackbar component with react-toastify notification in ProfileEditForm.tsx
Currently, our app uses the [Snackbar](https://github.com/OpenBeta/open-tacos/blob/eb13d0fe22a211f3201ccd9eee606504700f721a/src/components/ui/Snackbar.tsx#L3) component to display a message to the user when their profile is updated. This component is only used in the [ProfileEditForm.tsx](https://github.com/OpenBeta/open-tacos/blob/eb13d0fe22a211f3201ccd9eee606504700f721a/src/components/users/account/ProfileEditForm.tsx#L7) file.

I propose replacing the Snackbar component with the toast.info or toast.success function provided by the `react-toastify` library.

The updated ui would look something like this:
image

to something like
image

- Nice-to-have

hawkishpolicy:
Removed old Snackbar component.
Removed Snackbar div
Removed old justSubmitted and setJustSubmitted state
Added Toast.success for successful profile update
Added Toast.error for update failure
updated psuedo-code


src/components/users/account/ProfileEditForm.tsx
```diff
@@ -71,13 +70,15 @@ export default function ProfileEditForm (): ReactElement {
     if (profile != null) {
       // Update the profile object in the state
       setProfile(profile)
-      // Set the flag that indicates the user just submitted a change
-      setJustSubmitted(true)
+      // Removed old state for Snackbar
```
clintonlunn: Can you remove this reference to the Snackbar component? I don't think it's useful to know it used to be here.

clintonlunn (changes requested):

clintonlunn:
Thanks for the contribution. Just a few verbage changes I would suggest. If leaving a comment, it might be more useful to say what it does instead of "Added..[thing]", "Removed..[thing]".

clintonlunn:
@hawkishpolicy Sorry to go back on this, but I'm not sure how much those comments are adding, the things you wrote seem to pretty self explanatory and I would do minimal comments at most. Also, we may change the behavior in the future to move to the toast to a different position, so telling us the position of it currently might not make sense. We could just go find where we defined this toast initially to see what options it has. 

Also, can you check to make sure the Snackbar.tsx component is no longer used anywhere else, and then remove it and all references to it. We can use `react-toastify` completely in its place. 

Again, thanks for taking this on! Lemme know if you've got any other questions!

hawkishpolicy:
I'm happy to change it.  I'm brand new. Sorry for the kinks.

hawkishpolicy:
So can I just remove all comments and remove snackbar if it's no longer
being used elsewhere? that wasn't in the original bug post so i didn't want
to overstep.

I can do comments however you need them. It's unfortunately not very
intuitive to me about which comments are necessary and which are not.
Thanks for the guidance.

clintonlunn:
@hawkishpolicy You're right. It was a little bit late when I wrote up that issue :sweat_smile: 

when I said 

> I propose replacing the Snackbar component with the toast.info or toast.success function provided by the react-toastify library.

I meant removing Snackbar component entirely because it's not needed anymore, BUT I neglected to say that more explicitly.

These changes look good to me. @vnugent good to merge?

vnugent:
@all-contributors add @hawkishpolicy for code

PR #827, opened 2023-05-15
Username update error message

hawkishpolicy:
Changed username update .test error message to "User name cannot contain spaces"

### DIFF
--- a/src/components/lco/PageBanner.tsx
+++ b/src/components/lco/PageBanner.tsx
@@ -2,7 +2,8 @@
 import { OrganizationType } from '../../js/types'
 import Tooltip from '../../components/ui/Tooltip'
 import { InformationCircleIcon } from '@heroicons/react/20/solid'
-import { UsersIcon } from '@heroicons/react/24/outline'
+import { UsersIcon, ArrowUpRightIcon } from '@heroicons/react/24/outline'
+import { MobileDialog, DialogContent, DialogTrigger } from '../ui/MobileDialog'
 
 export interface LCOProfileType {
   /** Org unique id  */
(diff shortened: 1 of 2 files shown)


zichongkao:
Hey @hawkishpolicy! Thanks for looking into this! Not sure if you meant to continue working on this, but fyi the `checkUsername` function is somewhat more complicated than a no-space check. See https://github.com/OpenBeta/open-tacos/blob/65308a4c167ad67601ca2c24bf40ecbcd718349f/src/js/utils.ts#L163. Hence, simply saying it doesn't contain spaces isn't accurate. 

https://github.com/OpenBeta/open-tacos/pull/758#issuecomment-1527008004 would be helpful to reference. The two options here are: 
1. Change the error message to something like 'Username must be alphanumeric, optionally separated by a ".", "-" or "_".' which succinctly describes what `checkUsername` does, but could perhaps still be confusing.
2. Change the `checkUsername` regex to make it simpler, perhaps allowing alphanumeric and a few special characters (., _, -) and then updating the error message to say something like 'Username must be alphanumeric or ".", "-", "_".' 

What do you think?

clintonlunn:
@zichongkao can we also add @hawkishpolicy to the openbeta org so he can branch directly off of the project instead of a fork. It looks like in addition the the changes you mentioned. There are also old commits (before squashing and merging to develop) present in this branch. Idk how to do that.

zichongkao:
Yeah - that'd be for viet to handle!

hawkishpolicy:
Will do! I'll take a look and keep plugging away.

vnugent:
Added @hawkishpolicy to core-dev team

PR #823, opened 2023-05-13
[WIP] Add new organization edits to area query

zichongkao:

PR #829, opened 2023-05-16
Connect lco be and fe

actuallyyun:
- Add organizations to area query
- Pass organization query result to page banner component
- parse organization data to lco profile list

Questions for @vnugent  @zichongkao 
- Which fields in lco profile are required, and which aren't? For optional fields, what is the fallback state? 
- Can hand coded `data.tsx` be removed? 

@zichongkao I think this PR will have merge conflicts with yours [#823](https://github.com/OpenBeta/open-tacos/pull/823)

### DIFF
--- a/src/components/lco/PageBanner.tsx
+++ b/src/components/lco/PageBanner.tsx
@@ -1,6 +1,5 @@
 // import * as HoverCard from '@radix-ui/react-hover-card'
-import { ClimbType } from '../../js/types'
-import { LCO_LIST } from './data'
+import { OrganizationType } from '../../js/types'
 import Tooltip from '../../components/ui/Tooltip'
 import { InformationCircleIcon } from '@heroicons/react/20/solid'
 import { UsersIcon } from '@heroicons/react/24/outline'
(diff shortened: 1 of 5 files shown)


src/components/lco/PageBanner.tsx
```diff
@@ -12,38 +11,45 @@ export interface LCOProfileType {
...
-
-const isMyCrag = (ancestors: string[], myCrags: string[]): boolean =>
-  ancestors.some((path) => myCrags.some((myCragId) => myCragId === path))
-
-type PageBannerProps = Pick<ClimbType, 'ancestors'>
+interface PageBannerProps{orgs: OrganizationType[]}
+const getLcoList = (orgs): LCOProfileType[] => {
+  const lcoList: LCOProfileType[] = []
 
+  orgs.forEach(org => {
```
zichongkao: Might simplify the code a little to do map directly rather than push into an empty array:
```
const lcoList: LCOProfileType[] = orgs.map(org => ({
  id: ..
 }))
```

zichongkao: There orgs can have an orgType on them. While there is only one type now ('LOCAL_CLIMBING_ORGANIZATION'), there might be others in future so might be safer to filter for these here.

eg. `orgs.filter(org => org.orgType === 'LOCAL_CLIMBING_ORGANIZATION').map...`

src/js/graphql/gql/areaById.ts
```diff
@@ -47,6 +47,20 @@ export const QUERY_AREA_BY_ID = gql`
       }
       pathTokens  
       ancestors
+      organizations{
+        orgId
```
zichongkao, line 51: Then we would need to retrieve `orgType` here.

zichongkao (changes requested):
Looks good! Thanks for getting this out. One or two changes for your consideration. You checked that the FE works with this change?

src/components/lco/PageBanner.tsx
actuallyyun: Thanks for the suggestions! Good catch on checking orgType and using map.
Yeah currently it works on FE

src/js/graphql/gql/areaById.ts
zichongkao: Also remember to add a space after `organizations` and `content` before the curly braces. I guess our linter doesn't check graphql strings. 

An organization fragment might be nice at some point, but e can break that out later on when we need to reuse it.

src/components/lco/PageBanner.tsx
actuallyyun: Actually, since I have to change the keys of the org object, I am not sure how to do this with a map.

src/js/graphql/gql/areaById.ts
actuallyyun: true!

zichongkao (approved):
Looks good! Thanks for the changes :)

src/js/graphql/gql/areaById.ts
vnugent: is `organizations` field available for climb query as well?

vnugent (approved):

PR #839, opened 2023-05-22
Remove react-beautiful-dnd

zichongkao:
In response to https://github.com/OpenBeta/open-tacos/pull/696#discussion_r1199638229.

Took the more conservative approach of editing `CragTable` instead of deprecating. In the long run, `ClimbListPreview` should be able to replace it, but there are some vital differences that make it hard to just drop it in. Notably, `ClimbListPreview` receives data from react-hook-form context instead of from its props.

### DIFF
--- a/src/components/crag/cragTable.tsx
+++ b/src/components/crag/cragTable.tsx
@@ -6,7 +6,6 @@ import { getSetTypes } from '../ui/RouteTypeChips'
 import ButtonGroup from '../../components/ui/ButtonGroup'
 import { Button } from '../../components/ui/Button'
 import { summarize } from '../ui/Description'
-import { DragDropContext, Draggable, Droppable } from 'react-beautiful-dnd'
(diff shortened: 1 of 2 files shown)


src/components/crag/cragTable.tsx
```diff
@@ -199,51 +188,13 @@ export default function CragTable (props: CragTableProps): JSX.Element {
...
-                      <ClimbItem favs={favs} hideSummary={isEditing} key={i.id} climb={i} />
-                    </div>
-                  )}
-                </Draggable>
-              ))}
-            </div>
-          )}
-        </Droppable>
-      </DragDropContext>
+      <div className='grid grid-cols-1 lg:gap-3 xl:grid-cols-4 lg:grid-cols-3 md:grid-cols-2 fr-2'>
```
zichongkao, line 191: Stripped out the drag and drop functionality, leaving it just as a read-only list.

PR #840, opened 2023-05-23
Implement drag and drop for area climbs

zichongkao:
We made subareas drag-and-droppable in https://github.com/OpenBeta/open-tacos/pull/696.

This PR achieves the same for climbs in an area. The main work is to abstract out a new `DraggableTable` component which we then re-use for climbs and subareas.

Screen Recording 2023-05-22 at 7 16 50 PM


src/components/edit/AreaCRUD.tsx
```diff
@@ -100,53 +59,24 @@ export const AreaCRUD = ({ uuid: parentUuid, areaName: parentName, childAreas, e
...
-                </div>
-                )
-              : null}
-          </DragOverlay>
-        </DndContext>
+        <DraggableTable
+          rowIds={Array.from(areaStore.keys())}
+          editMode={editMode}
+          onDragEnd={onDragEnd}
+          renderRow={(areaId: string, idx: number) => {
```
zichongkao, line 66: A little bit inelegant, but I don't really know of a better design pattern. We need to pass in both the child component and the data to fill it with. One other idea is to pass in a list of childProps.
```
<DraggableTable ..>
  <AreaItem><AreaItem ...>
<DraggableTable ..>
```
But then we would have to maintain extra state here in the parent component to order the child components, which is worse than slightly ugly code IMO.

vnugent:
I haven't looked at this closely.  Quick question, does content in the CSV editor stay in sync?

zichongkao:
> I haven't looked at this closely. Quick question, does content in the CSV editor stay in sync?

Good call out -- it doesn't sync. I'll need to fix that.

vnugent:
> > I haven't looked at this closely. Quick question, does content in the CSV editor stay in sync?
> 
> Good call out -- it doesn't sync. I'll need to fix that.

if it's too complicated, I'd prioritize fixing other bugs since there's already an alternate way to order climbs using the bulk editor.

vnugent:
We let react-hook-form manage the form state.  The CSV editor manipulates the climb list here:

https://github.com/OpenBeta/open-tacos/blob/67adc08e6e40181b8fc51e38bf7f4eab97d21de9/src/components/editor/CsvEditor.tsx#L38

Let's postpone this improvement.  I think the climb edit page can use some UX improvement

vnugent:
Obsolete.

Issue #821, opened 2023-05-12
Openbeta community link points to a dead url


clintonlunn:
## Steps to Reproduce
[the community link](https://github.com/OpenBeta/open-tacos/blob/develop/CONTRIBUTING.md?plain=1#L28) within the contributing guidelines points to a dead link

## Expected Behavior
There is no longer a message board as it was not utilized much. This link can be removed. 

## Current Behavior
the link points to community.openbeta.io which returns 404
![image](https://github.com/OpenBeta/open-tacos/assets/24685932/2026744e-5e58-4ec1-8906-0f4b8cc6fb8c)

clintonlunn:
@vnugent do we use a message board or are we only on discord nowadays?

vnugent:
we retired the message board lat year due to low traffic.  I think we can remove the community link.

PR #850, opened 2023-05-30
Removed community link from CONTRIBUTING.md

josh610:
Per issue [#821](https://github.com/OpenBeta/open-tacos/issues/821). Only the Discord link is there now

### DIFF
--- a/src/js/graphql/gql/fragments.ts
+++ b/src/js/graphql/gql/fragments.ts
@@ -160,7 +160,6 @@ export const QUERY_ALL_COUNTRIES = gql`
       areaName
       uuid
       totalClimbs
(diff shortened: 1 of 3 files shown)


src/js/graphql/gql/fragments.ts
```diff
@@ -160,7 +160,6 @@ export const QUERY_ALL_COUNTRIES = gql`
       areaName
       uuid
       totalClimbs
-      updatedAt
```
vnugent, line 163: Can you create a new branch for each PR so that we don't have old (already merged) changes?  Thanks!

vnugent (changes requested):

PR #863, opened 2023-06-08
fix: don't use Next.js' Link to prevent prefetching

### LINKED ISSUE #860: Page not found after changing username
## Steps to Reproduce
1.  Log in
2. Go to profile -> Edit
3. Edit username and save
4. Being redirected back to the profile page but it's now a 404, not found

introduced by #838

vnugent:
Fix #860

### DIFF
--- a/src/components/users/account/UsernameChangeForm.tsx
+++ b/src/components/users/account/UsernameChangeForm.tsx
@@ -10,7 +10,6 @@ import { Input } from '../../ui/form'
 import { RulesType, Username } from '../../../js/types'
 import useUserProfileCmd from '../../../js/hooks/useUserProfileCmd'
 import Tooltip from '../../ui/Tooltip'
-import Link from 'next/link'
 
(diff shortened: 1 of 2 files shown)


src/components/users/account/UsernameChangeForm.tsx
```diff
@@ -149,9 +150,7 @@ export const UsernameChangeForm: React.FC = () => {
         ? (<h1>Create a username</h1>)
         : (
           <>
-            <Link href={`/u/${username}`}>
-              <a className='link flex gap-2 items-center'><ArrowLeftCircleIcon className='w-5 h-5' />Back to profile</a>
-            </Link>
+            <a className='link flex gap-2 items-center' href={`/u/${username}`}><ArrowLeftCircleIcon className='w-5 h-5' />Back to profile</a>
```
vnugent, line 153: Switch to regular `a` tag instead Next.js `Link`.  Using `Link` will cause Next to prefetch the new profile page `/u/<new_username` as a user typing a new name.

vnugent (commented):
.

PR #865, opened 2023-06-08
Small fixes to ensure dateClimbed as date works

zichongkao:
Tick can be added (This was before validation. Today is 6/8. Adding 7/8 shouldn't have been possible):
image

Date validation works:
image

### DIFF
--- a/src/components/users/TickCard.tsx
+++ b/src/components/users/TickCard.tsx
@@ -8,7 +8,7 @@ interface Props{
   ticks: TickType[]
   setTicks: Function
   tickId: string
-  dateClimbed: Date
+  dateClimbed: number
   notes: string
   style: string
 }
@@ -33,7 +33,7 @@ export default function TickCard ({ tickId, ticks, setTicks, dateClimbed, notes,
   return (
     <div className='flex flex-row justify-between px-3 py-3 mb-3 border-2 rounded-lg border-slate-100' id={tickId}>
       <div className='flex flex-col'>
-        <p className='text-md text-left text-gray-500'>{dateClimbed.toLocaleDateString()}</p>
+        <p className='text-md text-left text-gray-500'>{new Date(dateClimbed).toLocaleDateString()}</p>
         <p className='text-sm text-gray-500 mb-0'>{notes}</p>
       </div>
       <div className='flex flex-row'>
--- a/src/components/users/TickForm.tsx
+++ b/src/components/users/TickForm.tsx
@@ -24,7 +24,8 @@ const TickSchema = Yup.object().shape({
   attemptType: Yup.string()
     .required('Please choose an ascent type'),
   dateClimbed: Yup.date()
-    .required('Please include a date'),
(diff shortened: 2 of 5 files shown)


src/components/users/TickCard.tsx
```diff
@@ -8,7 +8,7 @@ interface Props{
   ticks: TickType[]
   setTicks: Function
   tickId: string
-  dateClimbed: Date
+  dateClimbed: number
```
zichongkao, line 11: The backend returns the date as a Unix timestamp.

src/components/users/TickForm.tsx
```diff
@@ -91,7 +92,7 @@ export default function TickForm ({ open, setOpen, setTicks, ticks, isTicked, cl
       userId: session.data?.user.metadata.uuid,
       style: style.name,
       attemptType: attemptType.name,
-      dateClimbed: dateClimbed,
+      dateClimbed: new Date(Date.parse(`${dateClimbed}T00:00:00`)), // Date.parse without timezone converts dateClimbed into local timezone.
```
zichongkao, line 95: The backend accepts anything that `new Date(...)` will accept, so we could pass in a number such as `Date.parse(..)` alone. However, the `Yup.Date()` validation is a bit easier to read if these are dates.

src/components/users/TickForm.tsx
```diff
@@ -118,9 +119,16 @@ export default function TickForm ({ open, setOpen, setTicks, ticks, isTicked, cl
         }
       })
       .catch((error) => {
```
zichongkao, line 121: Yup validations throw errors too, and these don't have `grapjhQLErrors` fields.

src/pages/api/user/ticks.ts
```diff
@@ -60,12 +79,12 @@ const handler: NextApiHandler<any> = async (req, res) => {
         ret.forEach((tick) => {
           const newTick: Tick = {
             name: tick.Route,
-            notes: tick.notes,
```
zichongkao, line 63: This is a bug -- the CSV I downloaded uses "Notes":
image

zichongkao:
I didn't test MP import locally. But given that we know the CSV inputs i.e.  dates in the format "2022-01-01", and the fact that `new Date('2022-01-01T00:00:00')` results in the correct input to our backend, I'm confident that that change won't cause problems.

src/components/users/TickForm.tsx
```diff
@@ -24,7 +24,8 @@ const TickSchema = Yup.object().shape({
   attemptType: Yup.string()
     .required('Please choose an ascent type'),
   dateClimbed: Yup.date()
-    .required('Please include a date'),
+    .required('Please include a date')
+    .max(new Date(), 'Please include a date in the past'),
```
musoke, line 28: Is the date check done in the user's local timezone?  Looks like it from the comment on L#95?

musoke (approved):

vnugent (approved):

Issue #846, opened 2023-05-30
Photos in popular tag not shown


actuallyyun:
## Steps to Reproduce

1.Go to `https://openbeta.io`
2.Click on **Popular** tab

## Screenshots
![Screenshot 2023-05-30 at 11 01 16](https://github.com/OpenBeta/open-tacos/assets/87448230/d418e28b-dd50-4859-8ff5-0200100020ac)

## Expected Behavior
Each card should have a photo

## Current Behavior
Cards are showing placeholders

## Browser & version
<---  Example: Firefox 102.7.0esr  -->

## Operating system
<--- Example:  MacOS 13.2.1 -->

Issue #444, opened 2022-07-20
Photos on Popular page don't load


CocoisBuggy:
![image](https://github.com/OpenBeta/open-tacos/assets/64557383/1769ac07-03f4-48e8-8776-769e6ea5adf0)

Currently when a user navigates to the [popular](https://openbeta.io/?v=explore) panel on the homepage they are greeted with a slightly prototype-ish U.I that doesn't load media for areas that actually DO have tagged media.

## Suggestion

1. Correct these components so that they show actual media linked to areas. This need not be a carousel of all tagged media, but should ideally prioritize primary relationship rather than an inherited one through children. This requires legwork on the back-end so for the moment simply showing media will be sufficient.
2. Areas that have no media should either have a component with no image placeholder, or should not show at all. My opinion is that if the area is indeed a popular one, it will have some media in its children or as a primary relation.

vnugent:
Is this still an issue?

Ali7040:
Kindly assign this issue to me.

Ali7040:
@vnugent What component should i look here for this issue? I'm a bit confused.

PR #871, opened 2023-06-11
images display problem #444 issue #1839

Ali7040:
solve the issue #444 #846


vnugent:
@Ali7040 the linting errors are still there.  Can. you please make sure pre-commit checks are run locally?

PR #874, opened 2023-06-15
Fix the issue #444

### LINKED ISSUE #444: Photos on Popular page don't load

Ali7040:
Fix #444


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
vnugent: I'm not sure why this was added?

vnugent:
The unit test was failing due to some HTML Canvas test snapshot changes.  I think it's ok to update the snapshot and push new commit.

```
yarn test -u
```

vnugent:
@Ali7040 I undid changes to yarn.lock and removed the extra / in the image path.  Thank you for your help!

---

Screenshot 2023-06-18 at 7 31 43 AM

Issue #879, opened 2023-06-19
Style Issue on popular page. card tags overlap with others.  


Ali7040:
## Screenshots
<img width="748" alt="Screenshot 2023-06-19 084937" src="https://github.com/OpenBeta/open-tacos/assets/81979505/6c9b6c7f-e3a1-4477-bb95-4457c1c1f257">

3.

## Expected Behavior

need padding between them.

## Current Behavior

card tags overlap

## Browser & version

chrome Version 114.0.5735.134

## Operating system
Window 10

PR #880, opened 2023-06-19
fix: add padding to discipline tags

Ali7040:
issue #879


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
vnugent, line 1: Can you remove this file: .yarnrc.yml?  It seems to cause unexpected changes to the yarn.lock file

vnugent (changes requested):
I just left a comment.  Can you fix linting errors?

.yarnrc.yml
Ali7040: okay.

vnugent:
Can you also fix linting errors?
https://github.com/OpenBeta/open-tacos/actions/runs/5330206526/jobs/9656652729?pr=880

vnugent:
can you also undo changes to `yarn.lock`?

PR #883, opened 2023-06-21
fix: add padding to discipline tags

Ali7040:
Issue #879


vnugent:
Screenshot 2023-06-21 at 10 27 31 AM

As you can see the spacing is not uniformed.  I think a better fix would be adding both horizontal and vertical padding (gap) to the parent flexbox

https://github.com/OpenBeta/open-tacos/pull/883/files#diff-1a7006bf6b524efdb131d8de63e12149cdb6b9bf319d193af8e967473434aa77R61

current
```html
<div className='my-1 text-xs flex flex-row space-x-2 flex-wrap'>{formatClimbingTypes(aggregate)}</div>
```

remove default flex-row and flex-wrap, add vertical & horizontal spacing
```html
<div className='my-1 text-xs flex gap-2'>{formatClimbingTypes(aggregate)}</div>
```

PR #898, opened 2023-07-02
Fix-897 Remove duplicate issue templates.

Ali7040:


actuallyyun:
Hey @Ali7040 Thanks for the pr. It seems that your pr contains commits from your other pr. 

Screenshot 2023-07-02 at 11 36 52

This might cause merge conflicts in the future with your other pr, also it's a good idea to keep the commits aligned with the problem your pr is trying to solve.

The problem might have been caused by checking out to a new branch(A) from a feature branch(B), and the new branch will have all the commits from branch B. 

There are different ways to fix it, I think the easiest way might be to cherry-pick your latest commit `3240346` to a new branch and create a new PR for it.  Please correct me if I'm wrong @vnugent

PR #908, opened 2023-07-11
fix: Searches for climbs should consider parent areas

### LINKED ISSUE #907: Search has inconsistent results
I am not sure how to title this one, feel free to edit if you have better ideas.  There is some chance that it reproducing it depends on the current state of the DB.

I was searching for https://openbeta.io/climbs/bbacb04a-98fe-50cc-9d2e-b2fac17290bd ("Robbers" on "Gem Boulder, 
The" at "Lynn Woods".)

I expected that "robbers lynn" would give a result quickly.  One of the things I like most about openbeta's search is that queries like this usually work.  As seen in the screenshot, it does not in this case.
image

I thought this might be due to "lynn" and "robbers" being too separated in the tree, but a similar search for "cops lynn" correctly finds a problem on the same boulder.
image

"robbers gem" does work:
image

"robbers massachusetts" also works:
image

musoke:
- [ ] refactor
- [x] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description

When searching for a climb, it is useful to be able to narrow the number of results by including the parent areas' names.  open-tacos does not include parent area names as a field when querying, so this did not work.

The data already exists in  typsense, so just tell open-tacos to use it.

### Related Issues

Fixes https://github.com/OpenBeta/open-tacos/issues/907

### What this PR achieves

Before:  the only result for query "robbers lynn woods" was "Home / USA / Colorado / Telluride / Norwood area / Ophir / Ophir Wall/Mirror Wall / Ophir Broke".  Presumably this was due to partial matches in the description: 
> **Lynn** and I went to Telluride to try and start a guide service for Royal **Robbins**,

image

After: exactly what I expected, although maybe not the optimal order.

image

### Notes

I haven't added any tests.

### DIFF
--- a/src/js/typesense/TypesenseClient.ts
+++ b/src/js/typesense/TypesenseClient.ts
@@ -68,7 +68,7 @@ export async function multiSearch (query: string): Promise<MultisearchReturnType
     searches: [
       {
         q: query,
-        query_by: 'climbName, climbDesc',
+        query_by: 'climbName, areaNames',
         collection: 'climbs',
         exclude_fields: 'climbDesc',
         page: 1,


vnugent:
what do you think if we drop `climbDesc` completely?   The intention at the time was it allows users to search for keywords in the description such as "crimpy face" or "offwidth".  The issue is it's not a well-known feature and most of the time we tend to search by name and location.  Having the description as a criteria throws off the results in some cases.

Example:
"Cat in the hat" is a popular climb name.  Let's say I want to search for "cat in the hat red rock"

- query: climbName, climbDesc, climbArea: 6th place [*]
- query: climbName, climbArea: 1st place

---

[*] While a neighbor climb, Rabbit's Aret while has a different name, its description contains a lot of article **the** s and also "Cat in that" therefore Typesense returns it as the 1st match.

Screenshot 2023-07-11 at 7 02 57 PM


Edit: in the future we can introduce a simple search language to allow searching by a specific field. Ex: `-d offwidth`

musoke:
Yes, that makes sense.  Searching full climb and area descriptions could be
provided separately if needed later.

src/js/typesense/TypesenseClient.ts
```diff
@@ -68,9 +68,8 @@ export async function multiSearch (query: string): Promise<MultisearchReturnType
     searches: [
       {
         q: query,
-        query_by: 'climbName, climbDesc',
+        query_by: 'climbName, areaNames',
         collection: 'climbs',
-        exclude_fields: 'climbDesc',
```
vnugent, line 73: We still want to exclude `climbDesc` so that it's not included in the result json (to keep the payload light)

Screenshot 2023-07-11 at 8 12 39 PM

vnugent (changes requested):

src/js/typesense/TypesenseClient.ts
musoke: Ah, yes!

vnugent (approved):

src/js/typesense/TypesenseClient.ts
musoke: fixed that now

Issue #787, opened 2023-04-20
update ts-standard to 12.0.0


musoke:
## Steps to Reproduce

1. clone https://github.com/OpenBeta/open-tacos/
2. yarn install
3. yarn lint

## Expected Behavior

package versions in package.json are mutually compatible. 

## Current Behavior

Get warning

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.5.0

YOUR TYPESCRIPT VERSION: 4.9.5

Please only submit bug reports when using the officially supported version.

=============
Done in 9.33s.
```

## Steps to fix

See also fix of similar issue in https://github.com/OpenBeta/openbeta-graphql/pull/265

- change https://github.com/OpenBeta/open-tacos/blob/5be4774db2b0261dfac8603e55749dc2d67ee42e/package.json#L111 to version 12.0.0
- `yarn install`
- add and run `yarn fix` script as in https://github.com/musoke/openbeta-graphql/blob/0b193c16c880127a15171ae39459a0f48350da11/package.json#L60
- fix lints that can't be fixed automatically 
- remove warning from readme: https://github.com/OpenBeta/open-tacos/blob/be2064a64974b9dbebea8f49dc98d275aa495289/README.md?plain=1#L125-L127

giri-madhan:
Hi @musoke, could you please assign this issue to me? I can work on making these changes. Thank you!

actuallyyun:
Amazing! I have been seeing this error as well, thanks for fixing it in the other repo! @musoke  I could open a PR to fix it in this repo if this sounds good to you guys. @vnugent

actuallyyun:
Hey @musoke after running `yarn fix`, I still got many linting errors, and most of them are about this rule:https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-misused-promises.md

![Screenshot 2023-04-27 at 16 51 59](https://user-images.githubusercontent.com/87448230/234901057-30040b8a-8a0d-4824-bfdb-105abcf7e1e9.png)

Should I fix this error or ignore this rule? 
There are also two errors about triple slashes :
```
[...]

PR #939, opened 2023-07-26
fixed Issue #787

Ali7040:
I just upgraded TypeScript using` yarn upgrade typescript@latest`. I didn't upgrade all dependencies because it caused test errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "tailwindcss": "^3.1.8",
     "ts-standard": "11.0.0",
-    "typescript": "4.9.5"
+    "typescript": "^5.1.6"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -116,7 +116,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "tailwindcss": "^3.1.8",
     "ts-standard": "11.0.0",
```
vnugent, line 118: We want to upgrade ts-standard, not Typescript (unless it's required by ts-standard).

```
yarn upgrade ts-standard@^12.0.2
```

vnugent (changes requested):

package.json
Ali7040: okay

Ali7040: image
@vnugent, if I upgrade to `ts-standard@^12.0.2`, we will need to make changes in a lot of files, and even after trying, I still get errors. What should I do?

vnugent: Some of them area valid warnings and we'll have to fix them one by one.  If you want  I can take over this PR and fix them.

Ali7040: I will try. Can you please tell me which area has valid warnings?

Ali7040: There's a problem:` <form onSubmit={handleSubmit(onSubmit)}>`. I am getting the error message Promise-returning function provided to the property where a void return was expected. in most of the files. However, I have solved other problems. Can you please guide me?

Ali7040: @vnugent we don't have eslint rules file at the root of the project.

Ali7040: image
@vnugent  Now only 8 warnings remain, and I have been trying to solve them, but I haven't found any solution yet.

vnugent:
superseded by https://github.com/OpenBeta/open-tacos/pull/970

PR #941, opened 2023-07-27
chore(docs):add Bader's info to openbeta-devs file

baderk:
---
name: Bader
about: walking through dev onboarding
title: ''
labels: ''
assignees: @actuallyyun 

---

- [ ] refactor
- [ ] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [x] other

## Description
walking through dev onboarding

### What this PR achieves
Adds my name as contributor.


src/assets/contributors-map.geojson
```diff
@@ -4,15 +4,15 @@
     {
       "type": "Feature",
       "properties": {
-        "firstName": "Yun",
```
vnugent: @baderk I think you need to add a new entry for your info, not updating the existing one (Yun)

vnugent (changes requested):

src/assets/contributors-map.geojson
baderk: Oh I see yes I will change that

baderk: Fixed the change here
3e08a7a94d65e11fd55c39bfc0b2bea91ef6079d

src/assets/contributors-map.geojson
```diff
@@ -30,6 +30,21 @@
...
+      "type": "Feature",
+      "properties": {
+        "firstName": "Bader",
+        "githubId": "baderk",
+        "favoriteCrag": "becfb41f-3278-59c6-a376-2d23d97c220d"
+      },
+      "geometry": {
+        "type": "Point",
+        "coordinates": [
+          36.037803,
```
vnugent: can you check if the coordinates are in [longitude, latitude] order?
Geojson specification: https://www.rfc-editor.org/rfc/rfc7946#section-3.1.1

If you can also use https://geojson.io/ to check the format of your file.  You should see drop pins if the format is correct.

vnugent (changes requested):

src/assets/contributors-map.geojson
baderk: Thanks for pointing out the specification and the geojson tool. I just addressed it and I see the 3 points now on the map 👍🏼

vnugent (approved):

vnugent:
@all-contributors add @baderk for code

PR #947, opened 2023-07-31
fix: Implementing 404 page for unknown crags and climbs

### LINKED ISSUE #927: 404 page for crags/routes that do not exist
**What would you like to be able to do? Please describe.**

When a user navigates to a page that does not exist (for example, https://openbeta.io/crag/79be0b81-140b-5006-b650-0b440c2e4d88 or https://openbeta.io/crag/79be0b8) they get an opaque error.
```
Application error: a client-side exception has occurred (see the browser console for more information).
```  
image

This should instead be a page making it clear that there is no error, the page just doesn't exist.   Maybe even flag malformed UUIDs if possible.

- Nice-to-have

sjsikora:
---
name: Pull request
about: Create a pull request
title: Implementing 404 page for unknown crags and climbs
labels: ''
assignees: ''

---

- [ ] refactor
- [x] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

Addressing Issue #927

### What this PR achieves

As pointed out in the issue, if you go to https://openbeta.io/crag/79be0b8 or https://openbeta.io/climbs/428f9f46-7dab, you are taken to a page that looks like this:

image

This is because we are trying to query the graphqlClient with an invalid UUID. This throws an unhandled GraphQLError error in getStaticProps functions of both the crag and climbs page. 

What this PR achieves is that it handles this error by returning { notFound: true } in getStaticProps, which consequently leads to the 404 page.

image

### Notes

 - I ensured that this error will only trigger when there is an invalid UUID. The idea being if there were other errors, there may be a different page. I wonder if this was the right approach or if a general catch-all would work.
 
- This is actually my first pull request, so I would appreciate some strict scrutiny of my code.

### DIFF
--- a/src/js/graphql/api.ts
+++ b/src/js/graphql/api.ts
@@ -215,6 +215,6 @@ export const getClimbById = async (id: string): Promise<ClimbType> => {
       id
     },
     fetchPolicy: 'no-cache'
-  })
+  }).catch(e => { throw new Error(e) })
   return res.data.climb
 }
--- a/src/pages/climbs/[id].tsx
+++ b/src/pages/climbs/[id].tsx
@@ -32,6 +32,7 @@ import { removeTypenameFromDisciplines } from '../../js/utils'
 import { TotalLengthInput } from '../../components/edit/form/TotalLengthInput'
 import { LegacyFAInput } from '../../components/edit/form/LegacyFAInput'
 import { getClimbById } from '../../js/graphql/api'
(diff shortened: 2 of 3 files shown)


vnugent (approved):

vnugent:
@all-contributors add @sjsikora for code

vnugent:
Thanks for your help.  A small suggestion for next time: please name your commit message per this convention https://www.conventionalcommits.org/en/v1.0.0/#summary

Issue #940, opened 2023-07-26
Crag name on the map is duplicated


baderk:
## Steps to Reproduce
https://openbeta.io/crag/edb2ca43-0cd8-512b-9ad4-7643b5cf9819
1. Click on a crag
2. Scroll down to the map
3. The name of the crag is in purple box with white font as well as a bigger name behind it in black.

## Screenshots
![Screenshot 2023-07-26 at 6 38 30 PM](https://github.com/OpenBeta/open-tacos/assets/10656279/a60d9626-afcc-479a-acc5-0b3cac272198)

## Expected Behavior
See one label instead of two.

## Current Behavior
Duplicate crag label

## Browser & version
Firefox 115.0.2 (64-bit)

## Operating system
Macos 13.4.1

sjsikora:
It looks like this comes from a broken feature.

From my assumption, the text in the back is supposed to correlate to the location of the main area of the page, and the violet tag would map to the children.

There is broken logic behind this code causing it to write the main name twice, but more importantly, it seems that the area response from the query does not have lat and long metadata of its children.

This area object is defined in /pages/crag/[id].tsx:

```
const rs = await graphqlClient.query<{ area: AreaType, getAreaHistory: ChangesetType[] }>({
    query: QUERY_AREA_BY_ID,
    variables: {
      uuid: params.id
    },
    fetchPolicy: 'no-cache'
  })
 ```
 and looking further into the object shows that the lat and long data is missing

![image](https://github.com/OpenBeta/open-tacos/assets/20007305/d36fc9cc-328a-46c3-b415-e9930c9f22e9)


Any ideas on how to get this data?

**Comment by vnugent** [comment]
@sjsikora 

GQL query https://github.com/OpenBeta/open-tacos/blob/develop/src/js/graphql/gql/areaById.ts#L96-L100

You can add lat, lng fields to the GQL query as follows
```gql
        metadata {
          leaf
          isBoulder
          leftRightIndex
          lat
          lng 
        }

```

sjsikora:
Thank you for the help! I will try to go in and repair this feature.

PR #953, opened 2023-08-02
fix: Fixed map for crags showing duplicate titles and not showing children on map

sjsikora:
---
name: Fixed map for crags showing duplicate titles and not showing children on map
about: 
title: ''
labels: ''
assignees:

---

- [ ] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

Issue #940

### What this PR achieves

As the discussion on the issue pointed out, the map is intended to label the main area of the page and label all the sub-areas on the map. Currently, it does not do that; it writes the main title twice on the map.

To fix this, I modified the query parameters (areaById) to send back the lat and long metadata of the sub-areas. Then, I fixed the map logic to write the sub-areas on the map and ensured the main area label was not written twice.

### Screenshots, recordings

Current:
image

After fix:
image

### Notes

The style of the labels may be want to be modified.

There also seems to be logic to change the color of labels based on variables "focused" and "selected." I left these, but they will not be used because the focus is set to null, and selected is the main area.id, which will never be equal to a subarea id.
```
            ${subArea.id === props.focused ? 'bg-green-500 text-white scale-125' : ' bg-white'}
            ${subArea.id === props.selected ? 'bg-violet-500 text-white scale-125' : ' bg-white'}`}
```

### DIFF
--- a/src/components/area/areaMap.tsx
+++ b/src/components/area/areaMap.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 import { Map, Marker, ScaleControl, LngLatBoundsLike, MapboxMap } from 'react-map-gl'
 import { AreaType } from '../../js/types'
 import { MAP_STYLES } from '../maps/BaseMap'
+import { Padding } from '@math.gl/web-mercator/dist/fit-bounds'
 
 interface Mappable {
   id: string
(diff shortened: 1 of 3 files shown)


vnugent (approved):

vnugent:
@sjsikora thanks so much for reviving the map feature!

Issue #751, opened 2023-03-23
Recent history cards display variable names rather than values


actuallyyun:
## Steps to Reproduce

1.Go to edit page [https://openbeta.io/edit](edit )
2.in the _Recent history_ component, the cards for each history entry are displaying variable names such as 
```
updatedBy.sub_type
updatedBy.buffer
updatedBy.position
metadata.isBoulder
metadata.leaf
metadata.lnglat.type
metadata.lnglat.coordinates[0]
metadata.lnglat.coordinates[1]
```

## Screenshots
![Screenshot 2023-03-23 at 13 26 14](https://user-images.githubusercontent.com/87448230/227203940-34703f2a-b7c6-4bbe-9423-dc3b52f97528.png)

## Expected Behavior

Maybe it should show meaningful values of these variables.

## Current Behavior

Currently, it does not provide meaningful information. I am not sure if this is intended. If not, I could try to fix it. 
## Browser & version
<---  Example: Firefox 102.7.0esr  -->
Chrome
## Operating system
<--- Example:  MacOS 13.2.1 -->

actuallyyun:
Currently, in the `history` query result, the `UpdatedFields` only contain fields and thus were shown on the history cards.
```
for instance: 
updatedFields
: 
Array(50)
0
: 
"_change.user.sub_type"
1
: 
"_change.user.buffer"
2
: 
"_change.user.position"
3
: 
"_change.historyId"
4
: 
"_change.prevHistoryId"
5
: 
"_change.operation"
6
: 
"_change.seq"
7
: 
"climbs[0].sub_type"
8
: 
"climbs[0].buffer"
9
: 
"climbs[0].position"
10
: 
"climbs[1].sub_type"
11
: 
"climbs[1].buffer"
12
: 
"climbs[1].position"
13
: 
"climbs[2].sub_type"
14
: 
"climbs[2].buffer"
15
: 
"climbs[2].position"
16
: 
"climbs[3].sub_type"
17
: 
"climbs[3].buffer"
18
: 
"climbs[3].position"
```
To fix this, it seems like it should be done in the db. What do you think?

vnugent:
Listing changed field names was a temporary solution.  Maybe we can display the field value instead?

andrew-jp:
I'll check this out

andrew-jp:
[...]

PR #954, opened 2023-08-02
History cards #751

andrew-jp:
name: Recent history cards #751 
---

- [ ] refactor
- [x] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description
 I updated the cards on the edit page to display more info - just the name of the changed attribute and the new value. This required making the 'FRAGMENT_CHANGE_HISTORY' query in contribs.ts larger. I also updated the margins on the cards themselves because they were inconsistent/nonexistent.
### Related Issues

Issue #751 

### What this PR achieves
Makes the cards on the /edit page look better, and they provide some info, rather than just the name of the fields. Hide the unnecessary (weird metadata) or inaccessible (i.e. 'children', which is an array of AreaType that I didn't include in the query).

### Screenshots, recordings
Screen Shot 2023-08-02 at 2 08 19 PM

### Notes
I had some issues with types and accessing objects - currently I don't have a type for the doc that's passed in. I first typed it as (ClimbType | AreaType | OrgType) but I needed the type name that it initially comes with (in types.ts it includes '... & DocumentTypeName') and couldn't work that out without removing the typing.

I'm open to any advice or changes.

### DIFF
--- a/src/components/edit/RecentChangeHistory.tsx
+++ b/src/components/edit/RecentChangeHistory.tsx
@@ -58,7 +58,6 @@ const ClimbChange = ({ changeId, fullDocument, updateDescription, dbOp }: Change
   if (fullDocument.__typename !== DocumentTypeName.Climb) {
     return null
   }
-  const doc = fullDocument as ClimbType
   // @ts-expect-error
   const icon = dbOpIcon[dbOp]
   return (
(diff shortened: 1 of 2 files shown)


vnugent:
@all-contributors add @andrew-jp for code & ideas

PR #974, opened 2023-09-20
Set up an initial SA grade context, so that routes can be added in th…

CocoisBuggy:
…

This is not a strictly accurate context to give, as there are bespoke SA systems for trad and sport. Nevertheless, this should get us off the ground with pretty close conversions

---
name: Pull request
about: Create a pull request
title: ''
labels: ''
assignees: ''

---

- [x] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### What this PR achieves

Not a permanent solution, but this will let me add South African routes in the interim. The provided context is not entirely accurate, and South Africa uses different grade contexts depending on the sub-area...

### Notes

This is not a scalable solution, a proper run at sandbag will need to be made at some point.

### DIFF
--- a/src/js/grades/Grade.ts
+++ b/src/js/grades/Grade.ts
@@ -36,6 +36,17 @@ const gradeContextToGradeScales = {
     aid: GradeScales.AID,
     snow: GradeScales.YDS, // is this the same as alpine?
     ice: GradeScales.WI
+  },
+  SA: {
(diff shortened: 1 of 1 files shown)


vnugent (approved):

vnugent:
Thanks @CocoisBuggy!

PR #976, opened 2023-09-22
Photogallerymodal tags

### LINKED ISSUE #645: Add tags to images within the PhotoGalleryModal
As a user, I want to see tags in images within the [PhotoGalleryModal](https://github.com/OpenBeta/open-tacos/blob/develop/src/components/media/PhotoGalleryModal.tsx), much like in pictures within a user profile page. 

An area with lots of photos: https://openbeta.io/crag/bea6bf11-de53-5046-a5b4-b89217b7e9bc

Current view (notice no tag overlayed onto the image):
image

view to emulate: if no climbs are tagged, only show the `+ new tag`, if climb(s) tagged, show the `+ new tag` and the new tag,
image
image

clintonlunn:
---
name: Pull request
about: Adding the taglist to the PhotoGalleryModal, add username to TagList

image

---

- [ ] refactor
- [x] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description

Refactored the previous `Tag` component into three separate components: `BaseTag`, `LocationTag`, and `UsernameTag`. In addition, there is a boolean flag to render the `UsernameTag` within the `TagList`. This can maybe be refactored in the future to include a list of tagtypes to include in the list. This results in the username tag only being included in certain parts of the app since we have username tags elsewhere in other parts of the app. 

### Related Issues

Issue #645

### DIFF
--- a/src/components/media/GalleryImageCard.tsx
+++ b/src/components/media/GalleryImageCard.tsx
@@ -1,0 +1,83 @@
+import { useState } from 'react'
+import Image from 'next/image'
+import clx from 'classnames'
+import Card from '../ui/Card/Card'
+import TagList from '../media/TagList'
+import { MobileLoader } from '../../js/sirv/util'
+import { MediaWithTags } from '../../js/types'
+import { getUploadDateSummary } from '../../js/utils'
+import { PostHeader } from '../home/Post'
+
+const MOBILE_IMAGE_MAX_WIDITH = 600
+
+interface GalleryImageCardProps {
+  header?: JSX.Element
+  mediaWithTags: MediaWithTags
+  onImageClick?: (event: React.MouseEvent<HTMLImageElement>, mediaWithTags: MediaWithTags) => void
+}
+
+/**
+ * Image card for the gallery page
+ */
+export const GalleryImageCard = ({
+  mediaWithTags,
+  onImageClick
+}: GalleryImageCardProps): JSX.Element => {
+  const [loaded, setLoaded] = useState(false)
+  const { mediaUrl, width, height, username } = mediaWithTags
+  const imageRatio = width / height
+  return (
+    <Card
+      header={<PostHeader username={username} />}
+      image={
+        <div className='relative block w-full h-full'>
+          <Image
+            src={MobileLoader({
+              src: mediaUrl,
+              width: MOBILE_IMAGE_MAX_WIDITH
+            })}
+            width={MOBILE_IMAGE_MAX_WIDITH}
+            height={MOBILE_IMAGE_MAX_WIDITH / imageRatio}
+            sizes='100vw'
+            objectFit='cover'
+            onLoad={() => setLoaded(true)}
+            className='rounded-md'
+            onClick={(event) => {
+              if (onImageClick != null) {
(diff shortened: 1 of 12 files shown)


clintonlunn:
@vnugent i'm really not sure why the test build step gets skipped. but if you pull down my branch, you'll see the failing tests

vnugent:
- I fixed the failing test and also relaxed some conditions in .github action file to make sure tests are run
- Rebased your working branch with the latest develop.  You should do pull to get the latest. Also run `yarn install` to pick up library upgrades.

clintonlunn:
> * I fixed the failing test and also relaxed some conditions in .github action file to make sure tests are run
> * Rebased your working branch with the latest develop.  You should do pull to get the latest. Also run `yarn install` to pick up library upgrades.

Thanks for the help. I think I see what you did here by mocking the child components to isolate just testing the PhotoMontage. I'll make sure I do this in the future!

vnugent:
Thank you for taking on this.  Some feedback:
1.  Area texts are hard to read.  Maybe use the default black text and set the background similar to how it's previously done in the user gallery?  (Neat and Cool tag)

Screenshot 2023-09-25 at 9 32 13 AM

2. It may be a bit cluttering to have full username next to tags . Can you use the person icon?  or does it make sense to reuse the [RecentMediaCard](https://github.com/OpenBeta/open-tacos/blob/develop/src/components/home/RecentMediaCard.tsx) so the gallery look consistent/similar to the front page? 

---
Screenshot 2023-09-25 at 9 18 09 AM

clintonlunn:
@vnugent I agree they are hard to read? 

I tested out putting the recent media cards into the photogallery. What do you think of this? I feel like we might want to put some kind of border around to separate each card. I also feel that way about the newTags page too, but would also defer to what you think

image

vnugent:
> @vnugent I agree they are hard to read?
> 
> I tested out putting the recent media cards into the photogallery. What do you think of this? I feel like we might want to put some kind of border around to separate each card. I also feel that way about the newTags page too, but would also defer to what you think
> 

I like the newer version.  Instead of adding borders, what about increasing the spacing between the cards?

>  I also feel that way about the newTags page too

which tag page?  the user profile?

clintonlunn:
image

@vnugent what do you think about this?

vnugent:
Can you please rebase? Not sure why photos are not loading in the preview build. We recently switched to the new image resizing service and introduced a PR that improves the way we load images.

src/components/media/GalleryImageCard.tsx
```diff
@@ -0,0 +1,83 @@
...
+interface GalleryImageCardProps {
+  header?: JSX.Element
+  mediaWithTags: MediaWithTags
+  onImageClick?: (event: React.MouseEvent<HTMLImageElement>, mediaWithTags: MediaWithTags) => void
+}
+
+/**
+ * Image card for the gallery page
+ */
+export const GalleryImageCard = ({
```
vnugent, line 22: Visually these cards are very similar to RecentMedia cards.  Let's merge after you rebase and refactor later.

vnugent:
Superceded by PR #1012

PR #1006, opened 2023-10-31
fix: add cooperative gesture to map

### LINKED ISSUE #993: Prevent map from zooming when scrolling area/climb screen
## Steps to Reproduce
1. Go to an area page such as https://openbeta.io/crag/bea6bf11-de53-5046-a5b4-b89217b7e9bc
2. Scroll to the end of the screen until the map is visible
3. At this point the map will capture the mouse scroll events (by the wheel or gesturing), interfering with the ability to scroll the to the top of the page

## Expected Behavior
I want to be able to scroll up and down.  This page has a good example https://docs.protomaps.com/
you  have to use Ctrl or ⌘ key to scroll

Also test the fix in mobile view to make sure when can pinch zoom and scroll

## Current Behavior
Hard to scroll when viewing the map

Ali-muhammed7040:


vnugent:
@all-contributors add @Ali-muhammed7040 for code

Issue #1013, opened 2023-11-05
Use new media servers


vnugent:
**What would you like to be able to do? Please describe.**
We now have 2 new media services, `media-stg.openbeta.io` and `media.openbeta.io`. Update the code to use the new url.

### Contributor To-do:
- [ ] In .env update  `NEXT_PUBLIC_CDN_URL` to **https://media-stg.openbeta.io**

### Admin to-do:
- [ ] Update NEXT_PUBLIC_CND_URL in Vercel dashboard to media.openbeta.io

Related issue: https://github.com/OpenBeta/open-tacos/issues/944

**How important is this to you** (Please pick one)
- Important

vnugent:
@ShravaniRoy this is a 1-line change task. Let me know if you're interested.

ShravaniRoy:
@vnugent yes, I am interested. Please assign it to me.

subhajit20:
hey @vnugent  , can I work on this?

PR #1019, opened 2023-11-09
Updated public CDN to new media server [  #1013]

enapupe:
- [x] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [x] optimization
- [ ] other

## Description
This pr replaces next's  and google storage cdn with our own image server. 
It also makes small tweaks in images loading with `MobileLoader`. 
We also remove `format=webp` because this should not be "enforced", browsers will tell the new image-server what format they accept. We go with webp > avif > jpg.

### Related Issues
Issue #1013

### What this PR achieves
Takes control of how images are optimized, also allowing us better cost/scalability control.

### Notes
I could not really understand why we have that `MobileLoader` messing with all image widths, something seems wrong there so I just replaced it and let the image width come naturally - it was set to 600px before, something seems missing.

### DIFF
--- a/.env
+++ b/.env
@@ -16,7 +16,7 @@ AUTH0_MGMT_CLIENT_SECRET=send request to hello at openbeta.io
 # Must prefix with NEXT_PUBLIC_in order to expose vars to the browser
 # See https://nextjs.org/docs/basic-features/environment-variables
 
-NEXT_PUBLIC_CDN_URL=https://storage.googleapis.com/openbeta-staging
+NEXT_PUBLIC_CDN_URL=https://stg-media.openbeta.io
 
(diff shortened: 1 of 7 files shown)


vnugent:
Can you see why there's an additional slash?  Other than that we're good to merge.
https://open-tacos-git-fork-enapupe-image-server-openbeta-dev.vercel.app/

Screenshot 2023-11-09 at 7 01 33 AM

enapupe:
Good catch. It has been fixed now.
I missed this one because I tested it before removing that `MobileLoader` from that component. 
As I mentioned above, I'm not sure if I understand why we had that mobileloader thing adding w=600 to all images. This PR removes it and let's `srcset` take care of loading the appropriate image.

Issue #881, opened 2023-06-20
Show neighboring climbs on climb page


josh610:
**What would you like to be able to do? Please describe.**
When looking at a climb, I like to be able to conveniently view other climbs at the same crag, à la Mountain Project. I think adding a list of these climbs to the climb page would encourage exploration of the crag.

Here's a quick mockup I created:
<img width="928" alt="Screen Shot 2023-06-20 at 12 47 36 PM" src="https://github.com/OpenBeta/open-tacos/assets/72105948/2974a817-b403-44df-b222-7d868b36e7b8">

**How important is this to you** (Please pick one)
- Important

vnugent:
Great idea!  Can you also create a mobile mockup as well?

josh610:
<img width="729" alt="Screen Shot 2023-06-27 at 2 22 15 PM" src="https://github.com/OpenBeta/open-tacos/assets/72105948/bcba0b45-2c42-43c3-b8b1-3c40b587d450">

I used a similar design to the one the MP app uses, which I think would suffice, but maybe someone with more experience in UI design could come up with something better.

vnugent:
Both mockups look great. Thanks!

clintonlunn:
is this feasible with the current data structure? i don't think we have access to all of the other climbs in an area when we load the climbs page? we do get the left and right climb and their left_right_index. See below for the object that gets returned when loading the page props for this route

```
{
    "pageProps": {
        "key": "05e7628c-cf5f-570e-bec1-98c6b31bc77c",
        "climb": {
            "id": "05e7628c-cf5f-570e-bec1-98c6b31bc77c",
            "uuid": "05e7628c-cf5f-570e-bec1-98c6b31bc77c",
            "name": "Golden Slumber",
            "fa": "Bob D'Antonio, Vaino Kodas, Moe Hershoff, 2002",
            "length": -1,
            "yds": "5.10d",
            "grades": {
                "ewbank": null,
                "font": null,
                "french": "6b+",
                "vscale": null,
                "yds": "5.10d",
                "uiaa": null,
                "__typename": "GradeType"
            },
            "safety": "UNSPECIFIED",
            "type": {
                "sport": true,
                "bouldering": false,
                "alpine": false,
                "tr": false,
                "trad": true,
                "mixed": false,
                "aid": false,
                "__typename": "ClimbType"
            },
            "media": [],
            "content": {
[...]

PR #1064, opened 2024-01-07
feat: add list of area's routes to route page

btmccord:
---
name: (feat: add list of area's routes to route page)
about: Route list on route page
title: ''
labels: ''
assignees: ''

---

- [ ] refactor
- [x] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

Issue #881 

### What this PR achieves

This PR adds a list of the areas routes to individual route pages 

### Screenshots, recordings

Screenshot 2024-01-06 at 21-46-57 Coyote Ugly · 5 10 · Sport

### DIFF
--- a/src/pages/climbs/[id].tsx
+++ b/src/pages/climbs/[id].tsx
@@ -297,7 +297,7 @@ const Body = ({ climb, leftClimb, rightClimb, parentArea }: ClimbPageProps): JSX
               </div>
             </div>
 
-            <div className='area-climb-page-summary-right lg:h-0 lg:min-h-full'>
+            <div className='area-climb-page-summary-right col-start-2 col-end-4 row-start-1 row-end-3'>
               <div className='mb-3 flex justify-between items-center'>
                 <h3>Description</h3>
               </div>
@@ -346,7 +346,7 @@ const Body = ({ climb, leftClimb, rightClimb, parentArea }: ClimbPageProps): JSX
                 {FormAction}
               </div>
             </div>
-            <div>
+            <div className='col-start-1 col-end-2'>
               <h4>Routes in {parentArea.areaName.includes(', The') ? 'The '.concat(parentArea.areaName.slice(0, -5)) : parentArea.areaName}</h4>
               <hr className='mt-2 mb-2 border-1 border-base-content' />
               {!editMode && <ClimbList gradeContext={parentArea.gradeContext} climbs={parentArea.climbs} areaMetadata={parentArea.metadata} editMode={editMode} routePageId={climbId} />}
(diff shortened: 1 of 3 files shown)


src/pages/climbs/[id].tsx
```diff
@@ -378,14 +385,18 @@ export const getStaticProps: GetStaticProps<ClimbPageProps, { id: string }> = as
     }
   }
 
-  const sortedClimbsInArea = await fetchSortedClimbsInArea(climb.ancestors[climb.ancestors.length - 1])
+  const parentAreaData = await getArea(climb.ancestors[climb.ancestors.length - 1])
```
Comment (vnugent) [inline src/pages/climbs/[id].tsx]: Good that you reuse getArea().  By default the call hits the API. Since we're not editing here and can afford cached data, can you add a 2nd param (for fetchPolicy), `cache-first`?

vnugent (changes requested):

src/pages/climbs/[id].tsx
```diff
@@ -295,7 +297,7 @@ const Body = ({ climb, leftClimb, rightClimb }: ClimbPageProps): JSX.Element =>
               </div>
             </div>
 
-            <div className='area-climb-page-summary-right'>
+            <div className='area-climb-page-summary-right lg:h-0 lg:min-h-full'>
```
Comment (vnugent) [inline src/pages/climbs/[id].tsx]: I had a close look at the css last night.  This addition looks a bit hackish if we can avoid it `lg:h-0 lg:min-h-full`, it would be great.  May I suggest the following?
1.  Put the new sibling climbs list JS in a reusable component (can be in the same file as \[id\].tsx)
2. Use tailwind breakpoints to show/hide that component

Example:

```html
<div className='area-climb-page-summary'>
  <div id='Title Information' className='area-climb-page-summary-left'>
    <!-- climb name, metadata, etc ... -->

   <div className='hidden lg:block'> <!-- For desktop: hide it by default, show it when screen >= lg  -->
      <SiblingClimbList {...propsEtc} />
   </div>
  </div>

  <div className='area-climb-page-summary-right'>
    <!-- right pane, description, etc-->
  </div>
</div>

<!-- for mobile -->
<div className='block lg:hidden'> <!-- the inverse of above, show it by default, hide it when screen  >= lg -->
  <SiblingClimbList {...propsEtc} />
</div>
```

vnugent (changes requested):

src/pages/climbs/[id].tsx
Comment (btmccord) [reply src/pages/climbs/[id].tsx]: We can actually accomplish this with the CSS Grid without duplicating the content. Submitting an commit with a cleaner solution using grid-area

Comment (vnugent) [reply src/pages/climbs/[id].tsx]: Great!  I file a refactoring task (lower priority) to consolidate/abstract away the the layout https://github.com/OpenBeta/open-tacos/issues/1066

vnugent:
@all-contributors add @btmccord for code and ideas

Issue #1078, opened 2024-01-15
feat: Climb/Route coordinates


btmccord:
**What would you like to be able to do? Please describe.**

Currently there is no way to set coordinates for climbs, only parent areas. Some areas are very large (or low density) and having the ability to give each climb coordinates would be helpful. 

It looks like this is already implemented in the back end,  

**How important is this to you** (Please pick one)

- Nice-to-have
- Important

clintonlunn:
If this is implemented, I think this would make it pretty straightforward to generate a map of ticked climbs and put it somewhere in the user's profile.

enapupe:
@clintonlunn I have a feeling this map would have just one marker per area if rendered from a state perspective, because there could be many in the same "pixel", so, not much would change then? I'm definitely not arguing against the coords in climbs itself.

alexoj46:
Hi! I would like to work on this.

Issue #1057, opened 2024-01-04
Area coordinates should be nullable and null should be default


musoke:
## Steps to Reproduce

1. Create a new area
2. Check the coordinates

## Expected Behavior

I expect that the coordinates for new areas are left unset unless there are known-good coordinates.  In addition, the coordinates should not default to those of the parent area.
It does make sense to display the map corresponding to a parent area when the coordinates are unknown.

## Current Behavior

When creating a new area, the app assumes its location is the same as the parent area.  This is not very helpful and is particularly frustrating once the assumed coordinates propagate down a few levels.

## Browser & version

All

## Operating system

All

vnugent:
We can leave newly created areas' [lat,lng] null therefore, inheriting their immediate ancestor's lat,lng.  However, things can get tricky when creating grand child areas.  Unless we update the db schema, we can't pass the parent's [lat,lng] down beyond its immediate children.

Perhaps, if you elaborate more on the user journey when creating new areas, we can make changes to the create area form?   Some options:
1. Introduce [lat,lng] field in create area form (pre-fill with the parent's values)
2. Make it easier to enter lat,lng eg introducing a coordinates picker, users can place a drop pin on the map.

musoke:
I am thinking about times when I create an area but don't know useful coordinates.  In these cases, adding coordinates is actively misleading.


Introducing an optional lat/long field in the create area form would be helpful.
Dropping a pin on a map would also be helpful.  However, both are separate from my issue here.

vnugent:
Ah I didn't think it through.  It makes sense for crags and boulders to have coordinates whereas parent areas are simply logical groupings similar to OSM relations.  Some immediate and longer term proposals:

Immediate:
1. Introduce lat/long field in the create area form
2. Update the map component: if the area is a crag/boulder then show the lat/long as a drop pin.  If it's a container area then show either its bbox or polygon, highlighting its child crags if available.

**Long term** (backend changes required)
Introduce a more comprehensive location/GIS object to area (#1051) 
   a. approach trails (can be gpx / reference to OSM trails)
   b. parkings and trail heads
   c. overview data (selected crags climbs to be highlighted in the area's overview map)

vnugent:
depends on https://github.com/OpenBeta/openbeta-graphql/issues/378

vnugent:
Backend PR https://github.com/OpenBeta/openbeta-graphql/pull/387 partially addresses this issue.  
- [x] Area lnglat field is nullable
[...]

PR #1092, opened 2024-02-09
feat: Climb/Route coordinates

alexoj46:
---
name: feat: Climb/Route coordinates

---

- [ ] refactor
- [X] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description

Implemented FrontEnd solution to Issue #1078 to add climb/route coordinates to climb pages.

### Related Issues

Issue #1078 

### What this PR achieves

Added FrontEnd component to climb/route page to display route coordinates. The component will also link to Google Maps, and will add a tick to the map on the website.

### Screenshots, recordings

Screenshot 2024-02-08 at 7 45 40 PM

### Notes
If coordinates are not present in metadata for that climb, the lat, lng display will still appear, with NA, NA as the value (as shown in the photo).

Also added my contributor information.


btmccord:
Rather than showing NA, we should inherit the parents coordinates.

musoke:
See https://github.com/OpenBeta/open-tacos/issues/1057 for discussion of
why inheriting parent coordinates may not be the best approach.

btmccord:
> See #1057 for discussion of why inheriting parent coordinates may not be the best approach.

That makes sense for areas, however if we have a coordinate for an area we should show it on the route page as a very high percentage of the time they will be that same (for our purposes as climbers).

Route coordinates are only really usefully for a handful of areas that are very low density.

What about indicating that the coordinates displayed are that of the area?

musoke:
Yes, that would probably do the trick - maybe even just show them grayed
out?

src/assets/contributors-map.geojson
```diff
@@ -136,6 +136,21 @@
...
+      "properties": {
+        "firstName": "Alex",
+        "githubId": "alexoj46",
+        "favoriteCrag": "b81f8942-4e04-56eb-a681-84568676c681"
+      },
+      "geometry": {
+        "type": "Point",
+        "coordinates": [
+          34.18008,
+          -85.81680
```
vnugent, line 150: Can you swap the order? Coordinate pair is expressed as longitude/lat (x,y) in geojson

vnugent:
I've just looked at the backend code.  When creating a new route, the backend populates the climb's coordinate pair with that of the parent.  We can hash out more detail in #1057.  For this PR we can just deal with the visual.  If the climb object doesn't have coordinates, it's not easy to fetch the parent's nor it is worth it to deal with in the frontend (I think the backend is a better place for it).  Can we just not show it if that's the case?

What about adding some wording to indicate the coordinates are inherited form the parent? Something like `34.18150, -85.81285 (Inherited)`?

---

Screenshot 2024-02-11 at 10 40 40 AM

btmccord:
> When creating a new route, the backend populates the climb's coordinate pair with that of the parent.

If this is the case I would think it's not possible to tell if it's inherited or not...?

I guess that right now they are all inherited since we have now editing functionality for climb coordinates currently?

vnugent:
> I guess that right now they are all inherited since we have now editing functionality for climb coordinates currently?

Right now they all inherited, but we can't edit climb coordinates at the moment.

vnugent (changes requested):

Issue #1107, opened 2024-02-26
bug: Page gets stuck on map when using scroll wheel


btmccord:
### Current Behavior
When scrolling a area page with the mouse scroll wheel the page stops scrolling (up or down) once you reach the map. At this point using the mouse wheel results in a "use crtl+scroll to zoom the map" overlay on the map

### Expected Behavior
I should still be able to scroll up and down the page with the mouse wheel when the map is on screen.

### Reproduction
-Go to an area page
-Scroll with a mouse wheel down to the map
-Try to scroll back up or continue down

vnugent:
The snapping css is probably defined incorrectly.

### Docs
- Mozilla https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scroll_snap
- Tailwind: https://tailwindcss.com/docs/scroll-snap-type

### Code
- https://github.com/OpenBeta/open-tacos/blob/597f7fb66a2b13a1a74a37e56f9298f6e3cc9749/src/app/(default)/layout.tsx#L30
- https://github.com/OpenBeta/open-tacos/blob/597f7fb66a2b13a1a74a37e56f9298f6e3cc9749/src/app/(default)/components/ui/AreaPageContainer.tsx#L29

btmccord:
This actually seems to only affect Firefox. Other browsers seem to ignore the snapping. What about just removing the snapping?

vnugent:
> This actually seems to only affect Firefox. Other browsers seem to ignore the snapping. What about just removing the snapping?

Sure. We can live without it.

PR #1109, opened 2024-03-01
fix: remove snapping on area page map

btmccord:
---
name: remove snapping from map on area pages
about: 
title: ''
labels: ''
assignees: ''

---

- [ ] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

Issue #1107 

### What this PR achieves

Removes snaping from the map element


vnugent (changes requested):
Can you also remove the entire classname in the global layout since it's no longer needed? Thanks!

`className='snap-proximity snap-y scroll-smooth'`

 https://github.com/OpenBeta/open-tacos/blob/develop/src/app/(default)/layout.tsx#L30

btmccord:
Yep, I wasn't sure is something else snapping relied on that..

Issue #1066, opened 2024-01-09
Refactor/improve AreaPageContainer to support 2-column layout


vnugent:
Both climb and area page use a 2-column layout defined by the following classes
- area-climb-page-summary
- area-climb-page-summary-left
- area-climb-page-summary-right

We should improve AreaPageContainer container to support left and right child component.

### Todo
- [ ] Improve AreaPageContainer to support a left and right child component
- [ ] Replace the divs in climb and area pages with the new component

- climb page: https://github.com/OpenBeta/open-tacos/blob/develop/src/pages/climbs/%5Bid%5D.tsx
- area page: https://github.com/OpenBeta/open-tacos/blob/develop/src/app/area/%5B%5B...slug%5D%5D/page.tsx

salgadopietrini:
Just confirming if this is still an open issue. Could I take on this one?

PR #1119, opened 2024-04-17
refactor summary sections

salgadopietrini:
---
name: Pull request
about: Create a pull request
title: ''
labels: ''
assignees: ''

---

- [x] refactor
- [ ] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

Issue #1066

### What this PR achieves

Currently, we have a common pattern where when we want to add a summary section (two columns), we create divs and add the classes that handle the grid layout. This PR encapsulates that logic into a single `<Summary />` component, which accepts two children (one for each column), and also implements it in places where the old logic was implemented:

- climb page 
- area page (through implementing it on `AreaPageContainer`)
- crag summary

### Screenshots, recordings
After refactor, should present no changes with current views:

- Area page, desktop
Screenshot 2024-04-17 at 18 01 57

- Climb page, desktop
Screenshot 2024-04-17 at 18 01 42

- Area page, mobile
Screenshot 2024-04-17 at 18 02 22

- Climb page, mobile 
Screenshot 2024-04-17 at 18 02 38

### Notes

- I saw that crag summary used the same pattern so I used the <Summary /> component there too. It should be pretty straightforward since the previous implementation was identical to the other section. However, I couldn't test changes on that section since I'm not very familiar yet so I don't how to reach that page.
- For each implementation, I extracted the component corresponding to the `left` and `right` columns into their own component. Let me know if this approach is ok or if you'd prefer that pass them directly to the props of `<Summary />`

### DIFF
--- a/src/app/(default)/components/ui/Summary.tsx
+++ b/src/app/(default)/components/ui/Summary.tsx
@@ -1,0 +1,15 @@
+import { ReactNode } from 'react'
+
+interface Props {
+  columns: {
+    left: ReactNode
+    right?: ReactNode
+  }
+}
+
+export const Summary: React.FC<Props> = ({ columns: { left, right } }) => (
+  <div className='area-summary'>
+    <div className='area-summary-left'>{left}</div>
+    {right !== undefined ? <div className='area-summary-right'>{right}</div> : null}
+  </div>
+)
--- a/src/app/(default)/components/ui/Summary.tsx
+++ b/src/app/(default)/components/ui/Summary.tsx
@@ -10,6 +10,6 @@ interface Props {
 export const Summary: React.FC<Props> = ({ columns: { left, right } }) => (
   <div className='area-summary'>
     <div className='area-summary-left'>{left}</div>
-    {right !== undefined ? <div className='area-summary-right'>{right}</div> : null}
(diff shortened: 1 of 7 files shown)


vnugent:
@salgadopietrini thanks for the PR.  I'll look into it this weekend.

src/app/(default)/components/ui/Summary.tsx
```diff
@@ -0,0 +1,15 @@
...
+  columns: {
+    left: ReactNode
+    right?: ReactNode
+  }
+}
+
+export const Summary: React.FC<Props> = ({ columns: { left, right } }) => (
+  <div className='area-summary'>
+    <div className='area-summary-left'>{left}</div>
+    {right !== undefined ? <div className='area-summary-right'>{right}</div> : null}
```
vnugent: can you test for undefined with `!=` to catch both undefined and null?  Ex: 
```javascript
right != undefined ?
```

vnugent (changes requested):

src/app/(default)/components/ui/Summary.tsx
salgadopietrini: yep, done!

vnugent:
@allcontributors add @salgadopietrini for code

PR #1125, opened 2024-04-26
feat: (#1113) save map view state

svkorepanov:
---
name: Pull request
about: Create a pull request
title: ''
labels: ''
assignees: ''

---

- [ ] refactor
- [x] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description

This pull request addresses https://github.com/OpenBeta/open-tacos/issues/1113, where we want to create a shareable link that we can communicate zoom level and lat long center values.

### Related Issues

Issue #1113 

### Screenshots, recordings

ezgif-7-e600230a0a

### DIFF
--- a/src/app/(maps)/components/FullScreenMap.tsx
+++ b/src/app/(maps)/components/FullScreenMap.tsx
@@ -1,11 +1,24 @@
 'use client'
-import { useEffect, useState } from 'react'
-import { GlobalMap } from '@/components/maps/GlobalMap'
+import { useCallback, useEffect, useState } from 'react'
+import { CameraInfo, GlobalMap } from '@/components/maps/GlobalMap'
+import { usePathname, useRouter, useSearchParams } from 'next/navigation'
 
 export const FullScreenMap: React.FC = () => {
   const [initialCenter, setInitialCenter] = useState<[number, number] | undefined>(undefined)
+  const [initialZoom, setInitialZoom] = useState<number | undefined>(undefined)
+  const router = useRouter()
+
+  const cameraParams = useCameraParams()
 
   useEffect(() => {
+    const initialStateFromUrl = cameraParams.fromUrl()
+
+    if (initialStateFromUrl != null) {
+      setInitialCenter([initialStateFromUrl.center.lng, initialStateFromUrl.center.lat])
+      setInitialZoom(initialStateFromUrl.zoom)
+      return
+    }
+
     getVisitorLocation().then((visitorLocation) => {
       if (visitorLocation != null) {
         setInitialCenter([visitorLocation.longitude, visitorLocation.latitude])
@@ -15,10 +28,18 @@ export const FullScreenMap: React.FC = () => {
     })
   }, [])
 
+  const handleCamerMovement = useCallback((camera: CameraInfo) => {
+    const url = cameraParams.toUrl(camera)
+
+    router.replace(url, { scroll: false })
+  }, [])
+
   return (
     <GlobalMap
       showFullscreenControl={false}
       initialCenter={initialCenter}
+      initialZoom={initialZoom}
+      onCameraMovement={handleCamerMovement}
     />
   )
 }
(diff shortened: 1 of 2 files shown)


musoke:
Would it be possible to format the coordinates in the URL similar to the OSM example?  Just so it's easier to read and to copy and paste.

I think it's just the "/" vs "%2F" that is different.  A comma might be even better.

src/app/(maps)/components/FullScreenMap.tsx
```diff
@@ -32,3 +53,47 @@ const getVisitorLocation = async (): Promise<{ longitude: number, latitude: numb
...
+    }
+
+    return queryToCameraInfo(cameraParams)
+  }
+
+  return { toUrl, fromUrl }
+}
+
+const cameraInfoToQuery = ({ zoom, center }: CameraInfo): string => {
+  return `${Math.ceil(zoom)}/${center.lat.toPrecision(3)}/${center.lng.toPrecision(3)}`
```
vnugent: - For latitude/longitude can you preserve at least 5 decimal places? 
 https://en.wikipedia.org/wiki/Decimal_degrees#Precision
- is it possible to keep the `/` in the url?

src/components/maps/GlobalMap.tsx
```diff
@@ -121,6 +144,7 @@ export const GlobalMap: React.FC<GlobalMapProps> = ({
         onDragStart={() => {
           setCursor('move')
         }}
+        onRender={onRender}
```
vnugent: I often see `onMove()` (instead of onRender) used to track current view state.  Would it work here?
https://visgl.github.io/react-map-gl/docs/get-started/state-management#controlled-map

vnugent (changes requested):

src/components/maps/GlobalMap.tsx
svkorepanov: 1. Done
2. It is not possible to do that using URLSearchParams, as it automaticaly encodes special characters.
I probably can prevent that by manually construction the parameter. However, it takes off any benefits that URLSearchParams gives.

Therefore, I believe it will be more beneficial to use URLSearchParams and put up with percent encoded (%2f) symbols in the url.

But if you're ok with keeping '/' slash in the url, you can check out the changes that I made

src/app/(maps)/components/FullScreenMap.tsx
svkorepanov: 1. Done
2. It is not possible to do that using URLSearchParams, as it automaticaly encodes special characters.
I probably can prevent that by manually construction the parameter. However, it takes off any benefits that URLSearchParams gives.

Therefore, I believe it will be more beneficial to use URLSearchParams and put up with percent encoded (%2f) symbols in the url.

But if you're ok with keeping '/' slash in the url, you can check out the changes that I made

src/components/maps/GlobalMap.tsx
svkorepanov: Oh, thank you. I was incorrectly reading documentation for [mapbox](https://github.com/mapbox/mapbox-gl-js) and  missed the fact that you're using maplibre. 

Fixed ✅

svkorepanov:
> Would it be possible to format the coordinates in the URL similar to the OSM example? Just so it's easier to read and to copy and paste.
> 
> I think it's just the "/" vs "%2F" that is different. A comma might be even better.

@musoke Thank you for the comment. I managed to format slash "/" without being percent encoded. Also gave my opinion on that [here](https://github.com/OpenBeta/open-tacos/pull/1125#discussion_r1582252269)

svkorepanov:
Updated footage

save-map

src/app/(maps)/components/FullScreenMap.tsx
```diff
@@ -1,11 +1,24 @@
...
+
+    if (initialStateFromUrl != null) {
+      setInitialCenter([initialStateFromUrl.center.lng, initialStateFromUrl.center.lat])
+      setInitialZoom(initialStateFromUrl.zoom)
+      return
+    }
+
     getVisitorLocation().then((visitorLocation) => {
       if (visitorLocation != null) {
         setInitialCenter([visitorLocation.longitude, visitorLocation.latitude])
```
vnugent, line 24: Preview build: https://open-tacos-git-fork-d-k-lippert-feat-save-map-url-openbeta-dev.vercel.app/maps

In the preview build deployed on Vercel.com, `visitorLocation` will be non-null. I'm getting 'camera=1/0.00000/0.00000' and the map is no longer initialized to my approximate location.  Perhaps adding a check 'when initialStateFromUrl == null **and** visitorLocation != null  then set URL param to visitorLocation?

vnugent (changes requested):

src/app/(maps)/components/FullScreenMap.tsx
svkorepanov: Fixed ✅

vnugent (approved):

vnugent:
@all-contributors add @svkorepanov for code

Issue #1137, opened 2024-05-23
About Page Icons Not Rendering


trevornei:
## Steps to Reproduce
Link: https://openbeta.io/about

1. Navigate from the home screen to the about page.
2. Scroll down to text sections for: Climbing data, How can I get involved?, why we need your donations?
3. Each section has an icon that is not rendering

## Screenshots
<img width="1440" alt="image" src="https://github.com/OpenBeta/open-tacos/assets/117675657/e28d91d3-ba5c-4797-9476-5166238fe99a">

## Expected Behavior
Should render icons.

## Current Behavior
Icons are not rendering.

## Browser & version
Chrome Version 124.0.6367.202 (Official Build) (arm64)

## Operating system
14.4.1 (23E224)

PR #1141, opened 2024-06-02
fix(about page): migrate to app dir, visual updates

vnugent:
---
name: Pull request
about: Create a pull request
title: ''
labels: ''
assignees: ''

---

- [x] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

Issue #1137


enapupe:
Looking sharp!

enapupe:
In any case, it seems the image loader was pointing static assets to the wrong place?

vnugent:
> In any case, it seems the image loader was pointing static assets to the wrong place?

I think so. Maybe when we introduced the custom image loader.  For the 2 images to load using normal relative path I have to add `Image.unoptimized` props.

https://nextjs.org/docs/app/api-reference/components/image#unoptimized

Issue #1145, opened 2024-06-06
Show existing and new coordinates when updating latitude and longitude in CoordinatePickerMap 


clintonlunn:
**What would you like to be able to do? Please describe.**

show existing coordinates
show new location. Maybe with a different icon (crosshair?)
reuse confirm dialog ([AlertDialogue](https://github.com/OpenBeta/open-tacos/blob/d1da49dc1bdb31d247bac1e6c25db346731af76d/src/components/ui/micro/AlertDialogue.tsx#L36))

![image](https://github.com/OpenBeta/open-tacos/assets/24685932/93f0255a-a571-421a-af22-a49408e9ab0c)

**How important is this to you** (Please pick one)
- Nice-to-have

clintonlunn:
@vnugent do you still want me to reuse the existing AlertDialogue component? The quick mockup only shows an 'OK' button to confirm.

vnugent:
Please reuse the alert dialog.  Initially I thought we should remove the full screen button, but I thought it's ok to leave it there.  Sorry for the confusion.

PR #1156, opened 2024-07-04
refactor: using alertdialog versus popup

clintonlunn:
---
name: Pull request
about: Create a pull request
title: ''
labels: ''
assignees: ''

---

- [x] refactor
- [ ] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description

Using the `AlertDialog` instead of a custom popup component. Display old and new coordinate with different `Marker` icons. 

### Related Issues

Issue #1145 

### Screenshots, recordings

image
image


clintonlunn:
@vnugent Looks like I might be getting the same errors as the last pr, preventing the preview from deploying. Not necessarily done with this pr or ready for review, but is there anything I can do to get the preview working?

src/components/maps/controls/CustomControl.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { useState, cloneElement } from 'react'
```
clintonlunn: I wanted to give an alternative approach to using the alertdialog also. I was curious about how to make a custom control and also use the control positions.

clintonlunn: @vnugent sorry I kind of let this pr sit for several weeks. Do you have any input on using a custom control versus the AlertDialog component like we had originally discussed? I'm fine either way, I have always been curious about how to make a maplibre custom control, so I explored doing that. 

in addition, I'm not sure why the preview is not working. I see the CI jobs are showing that docker-compose isn't found.

vnugent: I think you can use the existing AlertDialog.  We should do UI cleanup/overhaul at some point.
The Vercel preview works for me.  The docker-compose GitHub action may be broken. We can ignore it for now as long as Vercel build is green.

clintonlunn: @vnugent sounds good, I'll remove the custom control and use the AlertDialog component. 

In addition, I can't promise a time commitment (it's about to be climbing season 😄), but I would like to be involved in the ui cleanup effort.

vnugent:
@clintonlunn thanks for the fix.  I did a quick test notice the picker doesn't set initial location based on the form state:
1. From the area edit page, click on the coordinates
2. Pick a different location.  Click Confirm.
3. Click on the coordinate again.  Notice how the drop pin is set at the original location, not at what we set in step 2.

I think the fix may be having initializing the drop pin (in a `useEffect()` to the value in the form state

src/components/maps/CoordinatePickerMap.tsx
```diff
@@ -32,47 +32,39 @@ interface CoordinatePickerMapProps {
 export const CoordinatePickerMap: React.FC<CoordinatePickerMapProps> = ({
   showFullscreenControl = true, initialCenter, onCoordinateConfirmed
 }) => {
-  const [selectedCoord, setSelectedCoord] = useState({ lng: 0, lat: 0 })
-  const [cursor, setCursor] = useState<string>('default')
-  const [mapStyle, setMapStyle] = useState<string>(MAP_STYLES.light.style)
-  const [mapInstance, setMapInstance] = useState<MapInstance | null>(null)
-  const [popupOpen, setPopupOpen] = useState(false)
   const initialZoom = 14
+  // const [viewState, setViewState] = useState<Pick<ViewState, 'longitude' | 'latitude' | 'zoom'>>({
```
vnugent: can you also remove all the comments?

vnugent (changes requested):

clintonlunn:
> @clintonlunn thanks for the fix. I did a quick test notice the picker doesn't set initial location based on the form state:
> 
> 1. From the area edit page, click on the coordinates
> 2. Pick a different location.  Click Confirm.
> 3. Click on the coordinate again.  Notice how the drop pin is set at the original location, not at what we set in step 2.
> 
> I think the fix may be having initializing the drop pin (in a `useEffect()` to the value in the form state

@vnugent 
Thanks, that was it. Instead of passing an initialCenter prop to the coordinate map at all, we can just watch the form context for the `latlngStr` value and decide where to put the map pin.

clintonlunn:
@vnugent i refactored a bit to simplify this pr a lot. i believe all of the issues should be resolved now. please let me know if there is anything else you would like to see done

vnugent (approved):

PR #1168, opened 2024-08-13
Add Germany to international ToC

pamolloy:
Add Germany to the International ToC now that there is at least one area with routes! :smile:


vnugent:
@all-contributor add @pamolloy for code, ideas, content

PR #1170, opened 2024-09-14
fix: make fullName/display name an empty string if null

### LINKED ISSUE #1160: null shown in title on profile page
## Steps to Reproduce

1. Navigate to profile page without display name e.g. https://openbeta.io/u/jeff

## Expected Behavior
Page title in browser tab should show "jeff (/u/jeff)"

## Current Behavior
Page title in browser tab shows "null (/u/jeff)"

## Browser & version
Firefox 127

## Operating system
MacOS 14.5

clintonlunn:
---
name: Pull request
about: Create a pull request
title: ''
labels: ''
assignees: ''

---

- [ ] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description

If a user created an account, set a username but not a display name, the display name (fullName) would be `null`. To prevent this, we can use a nullish coalesce operator to check for null and convert to an empty string. 

NOTE: this did not happen if the user set a display name and then deleted it. In that case, the display name would already be an empty string, not reverted back to null. 

### Related Issues

Issue #1160

### What this PR achieves

Prevent the user from seeing `null` in the tab title. Instead of null, it will become an empty string. 

### Screenshots, recordings

(staging env without change)
image

(local env with change) 
image

### DIFF
--- a/src/js/hooks/seo/useUserProfileSeo.ts
+++ b/src/js/hooks/seo/useUserProfileSeo.ts
@@ -15,7 +15,7 @@ interface SeoTagsHookProps {
 export const useUserProfileSeo = ({ username = '', fullName = '', imageList = [] }: SeoTagsHookProps): SeoHookType => {
   const author = `/u/${username}`
   const count = imageList?.length ?? 0
-  const photoCountStr = `${count === 0 ? '' : count} Photo${count > 1 ? 's' : ''}`
+  const photoCountStr = `${count} Photo${count !== 1 ? 's' : ''}`
   const pageTitle = `${fullName ?? ''} (${author}) •  ${photoCountStr} on OpenBeta`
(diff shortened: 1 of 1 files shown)


clintonlunn:
@vnugent 

small bug fix. I defaulted the display name in the title to be an empty string, but the bug ticket indicates that it should instead be the username. Do you have a preference? 

i.e. `(/u/ntest) • 2 Photos on OpenBeta` versus `ntest (/u/ntest) • 2 Photos on OpenBeta`

clintonlunn:
Also, the wording was weird when the user had 0 photos uploaded, I fixed the grammar for that. 

Instead of saying. "Photo on OpenBeta" it now says "0 Photos on OpenBeta" 

image

PR #1173, opened 2024-10-03
fix: raise area description character limit from 3500 to 10000

julianlam:


vnugent:
@all-contributors add @julianlam for code and ideas

PR #1181, opened 2024-10-21
Fix gallery close button being hidden behind the title

julianlam:
Current behaviour on openbeta.io:

image

The close button is visible, and not clickable.

Noting that alternating the order of elements resolves the overlap issue.

Seems related (regression?) from fdb64aa049e37ee0128942d0e21ca3982e1d28ac which ironically was trying to fix the same thing.

## Reproduction Steps

1. Browse to a climb (e.g. [_Super Solenoid_](https://openbeta.io/climbs/2a25916c-152e-46d2-8b50-9fc909be17ec)), click on the uploaded image.
2. Note that the close button is hidden behind the header.

## Devices, etc.

* Fedora Linux, Firefox and Chromium — Reproduces
* Android, Firefox and Chrome — reproduces

### DIFF
--- a/src/components/ui/MobileDialog.tsx
+++ b/src/components/ui/MobileDialog.tsx
@@ -23,14 +23,14 @@ export const DialogContent = React.forwardRef<any, Props>(
           className={clx(fullScreen ? 'dialog-wide' : 'dialog-default')} {...props} ref={forwardedRef}
         >
           <div className='flex items-center px-2'>
+            <DialogPrimitive.Title className='dialog-title'>
+              {title}
+            </DialogPrimitive.Title>
             <DialogPrimitive.Close aria-label='Close' asChild className='dialog-close-button'>
               <button className='btn btn-circle btn-ghost outline-none'>
                 <XMarkIcon size={24} />
               </button>
             </DialogPrimitive.Close>
-            <DialogPrimitive.Title className='dialog-title'>
-              {title}
-            </DialogPrimitive.Title>
           </div>
           {children}
         </DialogPrimitive.Content>


clintonlunn:
Hi @julianlam I am not able to reproduce this issue. Can add some steps to reproduce this and include the some details on device and browser? 

@vnugent am I able to add @julianlam to the openbeta organization (or whatever capacity it requires) so he doesn't need to PR from a fork?

vnugent:
> am I able to add @julianlam to the openbeta organization (or whatever capacity it requires) so he doesn't need to PR from a fork?

I just added him

julianlam:
@clintonlunn Done

clintonlunn:
@julianlam after some investigation, I believe the issue is actually that the `<PhotoGalleryModal />` component is using the defaults.css, versus what was being tested in https://github.com/OpenBeta/open-tacos/commit/fdb64aa049e37ee0128942d0e21ca3982e1d28ac, that was referencing the global.css file. 

I think some evidence to point to this is the new /climb//climbId page when updated to Next13 in https://github.com/OpenBeta/open-tacos/pull/1195, the DialogClose button is correctly showing. 

So while your fix works, maybe the fix is instead to alter the css in defaults.css to match `global.css` so we aren't relying n order, but instead restricting the title from overlapping with the close btuton: 

```
  .dialog-title {
    @apply flex-grow text-center bg-base-100 leading-none;
  }

  .dialog-close-button {
    @apply mx-auto z-auto;
  }
```

What do you think?

src/components/ui/MobileDialog.tsx
```diff
@@ -23,14 +23,14 @@ export const DialogContent = React.forwardRef<any, Props>(
           className={clx(fullScreen ? 'dialog-wide' : 'dialog-default')} {...props} ref={forwardedRef}
         >
           <div className='flex items-center px-2'>
+            <DialogPrimitive.Title className='dialog-title'>
```
clintonlunn, line 26: @julianlam after some investigation, I believe the issue is actually that the `<PhotoGalleryModal />` component is using the `defaults.css` css rules, versus what was being tested in https://github.com/OpenBeta/open-tacos/commit/fdb64aa049e37ee0128942d0e21ca3982e1d28ac, that was referencing the global.css file. 

I think some evidence to point to this is the new /climb//climbId page when updated to Next13 in https://github.com/OpenBeta/open-tacos/pull/1195, the DialogClose button is correctly showing. This new page would be referencing `global.css`.

So while your fix works, maybe the fix is instead to alter the css in `defaults.css` to match `global.css` so we aren't relying on order, but instead restricting the title from overlapping with the close button: 

```
  .dialog-title {
    @apply flex-grow text-center bg-base-100 leading-none;
  }

  .dialog-close-button {
    @apply mx-auto z-auto;
  }
```

What do you think?

clintonlunn (changes requested):

src/components/ui/MobileDialog.tsx
julianlam: Given that the PhotoGalleryModal is only invoked by the climbs page (I don't actually know if this is true), and that it is being refactored, might be this PR is moot and can be closed :smile:

clintonlunn: For sure, I do think it's still valid to fix though because it definitely is an issue during the transition period of old climb page to new climb page.

mikeschen: Is there anywhere in the app now that we use the old climbs page? (https://openbeta.io/climbs/44c73a5e-9f32-599f-9e3a-37cf9d7b7fd4)

mikeschen:
Closing PR as we are removing the old Climbs page and using new next.js 13 Climbs page with Edit https://github.com/OpenBeta/open-tacos/pull/1307

PR #1193, opened 2024-10-31
fix: #1130, use gradesObj in TickButton instead of deprecated `yds` property

### LINKED ISSUE #1130: Recently added climb is not allowing me to tick climb (occurring across multiple climbs)
## Steps to Reproduce

1. Go to Hercules Hug in Washington state
2. attempt to tick climb
3. "cannot fetch climb grade" error appears, no tick is made

## Expected Behavior

should tick like any other boulder

## Current Behavior

does not tick climb and receive error message that grade can't be retrieved

## Browser & version
google chrome

## Operating system
windows

julianlam:
## Observed in Production

* Certain climbs are un-tickable. Attempting to tick them causes the TickForm to display the following error (image courtesy of @emilyg1709):
    * 
* It was also observed that some climbs are tickable, namely older ones in the United States.

## Reproduction in staging

* I was unable to find a suitable climb in staging that reproduced the issue, until @musoke suggested I create my own climb. Sure enough, the issue reproduced, which suggests that this occurs with newer climbs.

## The fix

* @daneashman commented in #1130 that a data issue may be the culprit. Affected climbs do not have a `yds` property, while unaffected climbs do.
* TypeScript hinting in code stated that `yds` is deprecated
* The `TickForm` component is dynamically built when the `TickButton` component is clicked.
* The `TickButton` component is rendered in the ClimbData page, and passes in `grade={yds}`, which is `null` on newer climbs
* I noticed that places where the grade was shown now used the `Grade` object(?) which had a `.toString()` method that worked on both Sport/Trad climbs and boulders. I updated the TickButton component to pass in `gradesObj.toString()` instead of `yds`
* Ticking the affected climb on staging now worked.

### DIFF
--- a/src/pages/climbs/[id].tsx
+++ b/src/pages/climbs/[id].tsx
@@ -440,7 +440,7 @@ const ClimbData: React.FC<{
 
         {!editMode && (
           <div className='mt-8'>
-            <TickButton climbId={climbId} name={name} grade={yds} />
+            <TickButton climbId={climbId} name={name} grade={gradesObj.toStringTradSportAid()} />
           </div>
         )}
       </div>
--- a/src/pages/climbs/[id].tsx
+++ b/src/pages/climbs/[id].tsx
@@ -440,7 +440,7 @@ const ClimbData: React.FC<{
 
         {!editMode && (
           <div className='mt-8'>
-            <TickButton climbId={climbId} name={name} grade={gradesObj.toStringTradSportAid()} />
+            <TickButton climbId={climbId} name={name} grade={gradesObj.toString()} />
           </div>
         )}
       </div>


clintonlunn:
@julianlam Would you mind filling in some details in this pr to include reproduction that you've done in the staging env and maybe some screenshots? 

@vnugent could you authorize a vercel deployment here?

julianlam:
Hi @clintonlunn — done!

It's worth noting here that my use of `gradesObj.toString()` is not standard. Other places in the codebase have the entire form revamped to use `gradesObj` entirely, so it could be that this is a band-aid until the `TickButton`/`TickForm` are refactored.

However, it is also worth noting that it seems odd to pass in all this extra data to the backend (specifically, `name` and `grade`) as those details could be retrieved server-side. I don't have enough expertise in this codebase to ask questions, though, so I'll leave it :smile:

julianlam:
@clintonlunn on staging the affected climb is _Discovery_, addressable via `climbs/ad8361ab-94cb-48fe-aa84-48f6a6ba7d12`

clintonlunn:
Yep, you can see that the `yds` value is undefined in new climbs but defined in older climbs. 

Looks like submitting with an older climb has the grade correctly populated.
image


It threw me a little that the `.toString()` was not this [.toString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString). The [gradesObj.toString()](https://github.com/OpenBeta/open-tacos/blob/62e326a1f19b68bbda02debcb0bf61f8e9441a4d/src/js/grades/Grade.ts#L127-L155) is pulling out the proper grade from the gradesObj

This looks good to me and unblocks a lot of people! @vnugent 



It does seem like the yds property is deprecated and should be removed in a future issue. It's imported into a few components but doesn't seem to be doing anything.

julianlam:
@vnugent May I get a PR review so this can be merged? :pray:

Vichy97 (approved):

clintonlunn (approved):
Looks good to me. In the future we should see if we can refactor to avoid needing to put a .toString() on this prop.

PR #1200, opened 2024-11-07
fix: og:image tags for /area/ and /climbs/

### LINKED ISSUE #1204: Image not loading in share preview for Climb and Area pages
## Steps to Reproduce

1.
2.
3.

## Screenshots
image
image

## Browser & version

## Operating system

julianlam:
- [ ] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description

- og:image tags for areas do not work (invalid URL)
- ~og:image tags for climbs do not work (shows default south africa image)~
    - Same issue as below, router is called twice, but `og:image` is not properly retrieved the first time due to null uid
- ~og:image tags for profiles do not work (arguments are double escaped)~
    - Upstream issue: https://github.com/facebook/react/issues/13838
    - Passing in restricted characters like `&` cause them to be escaped. There are workarounds but none that look simple (or maintainable) enough to implement.
    - When a user profile is loaded, `useUserProfileSeo` is called to load the photos for `og:image`
    - This is called by the react router, but the router is called **twice**. Once with a `null` uid (causing the `og:image` tag to be the default image), and a second time in-client causing the tags to be properly set in DOM. However, that is too late for link-preview services, because they don't parse js.

### What this PR achieves

Fixed the above issues.

### Notes

Query parameters are being escaped but I can't work around them. e.g. `?w=600&q=75` has the `&` replaced with `&amp;` which isn't read correctly by the image server. (full quality image is sent)

### DIFF
--- a/src/app/(default)/area/[[...slug]]/page.tsx
+++ b/src/app/(default)/area/[[...slug]]/page.tsx
@@ -199,7 +199,7 @@ export async function generateMetadata ({ params }: PageWithCatchAllUuidProps): 
 
   const name = sanitizeName(areaName)
 
-  const previewImage = media.length > 0 ? `${CLIENT_CONFIG.CDN_BASE_URL}/${media[0].mediaUrl}?w=1200q=75` : null
+  const previewImage = media.length > 0 ? `${CLIENT_CONFIG.CDN_BASE_URL}${media[0].mediaUrl}?w=1200&q=75` : null
 
   const description = `Community knowledge • ${wall}${name}`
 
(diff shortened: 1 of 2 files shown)


clintonlunn:
@julianlam just to point out I think the change you made to useClimbSeo only applies to the (soon to be deprecated/removed) /climbs/ page. The new /climb/ page does not have a generateMetadata() function yet, but should be added (out of scope for this issue). it is inheriting from https://github.com/OpenBeta/open-tacos/blob/4f4c7bac00122ff67e7e64e4cea812a7d8ae615f/src/app/(default)/layout.tsx#L13-L21

julianlam:
@clintonlunn Correct, I realized that earlier, which is why I made the commit. There still is an extra slash there, so having it removed is better, as the image 404s otherwise.

clintonlunn:
image
image

clintonlunn (approved):
This looks good to me. There are some discrepancies between the image url generation in `useClimbSeo()` (`/climbs/` page) and the metadata for `/area/` page maybe we can improve on that in the future since we're still seeing errors about an incorrect aspect ratio.

PR #1227, opened 2024-11-17
fix: should use uuid not username to check whether a user exists in db

vnugent:
- [ ] refactor
- [ ] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

Issue  #1224 

### What this PR achieves
Fix erroneous username check in PR #1224 causing the username to be overridden by the default


src/js/auth/initializeUserInDb.ts
```diff
@@ -16,10 +16,13 @@ interface InitializeUserInDBParams extends UpdateUsernameInput {
 
 const serverClient = getClient()
 
+/**
+ * Look up in our db (not Auth0) to see whether a user by uuid exists.  If it doesn't then insert a new user profile.
+ */
 export const initializeUserInDB = async (params: InitializeUserInDBParams): Promise<boolean> => {
   const { auth0UserId, accessToken, userUuid, username, email, avatar } = params
-  const existed = await doesUserExist(username)
-  if (existed) {
```
vnugent, line 22: `existed` will be false if the user has changed their username to something other than the initial default in Auth0. This causes their new username to be set to the default.

Issue #1250, opened 2025-01-08
Allow editing of profile image


Vichy97:
I see that some people have a profile image but I cannot find a way to actually edit this. We should support uploading a profile image

PR #1258, opened 2025-01-22
Allow editing of profile image

mikeschen:
---
name: Pull request
about: Create a pull request
title: Allow editing of profile image #1250
labels: ''
assignees:  Michael Schennum

---

- [ ] refactor
- [x] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description
Allow editing of profile image from the Edit Profile page
### Related Issues

Issue #
1250

### What this PR achieves

Allow editing of profile image from the Edit Profile page

### Screenshots, recordings

Screenshot 2025-01-22 at 12 06 37 PM

Screenshot 2025-01-22 at 12 06 49 PM

### DIFF
--- a/src/components/users/account/UpdateProfileForm.tsx
+++ b/src/components/users/account/UpdateProfileForm.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react'
+import { useEffect } from 'react'
 import { useForm, FormProvider } from 'react-hook-form'
 import { useSession, signIn } from 'next-auth/react'
 import { toast } from 'react-toastify'
@@ -9,7 +9,8 @@ import { PencilIcon } from '@heroicons/react/24/solid'
 import { Input, TextArea } from '../../ui/form'
 import useUserProfileCmd from '../../../js/hooks/useUserProfileCmd'
 import { BaseProfilePhotoUploader } from '../../media/BaseUploader'
-import { getProfilePhotoSrc } from '../PublicProfile'
+import { ProfileImage } from '../PublicProfile'
+import { useUserGalleryStore } from '../../../js/stores/useUserGalleryStore'
 
 const validationSchema = z
   .object({
@@ -46,7 +47,8 @@ export const UpdateProfileForm: React.FC = () => {
 
   const userUuid = session.data?.user.metadata.uuid
 
-  const [avatar, setAvatar] = useState<string | null>(null)
+  const setAvatarUrl = useUserGalleryStore(store => store.setAvatarUrl)
+  const avatar = useUserGalleryStore(store => store.avatarUrl)
 
   useEffect(() => {
     if (userUuid != null) {
@@ -55,7 +57,9 @@ export const UpdateProfileForm: React.FC = () => {
         if (profile != null) {
           const { displayName, bio, website, avatar } = profile
           reset({ displayName, bio, website })
-          setAvatar(avatar ?? null)
+          if (avatar != null) {
+            setAvatarUrl(avatar)
+          }
         }
       }
       void doAsync()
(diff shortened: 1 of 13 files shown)


src/components/users/account/UpdateProfileForm.tsx
```diff
@@ -4,9 +4,13 @@ import { useSession, signIn } from 'next-auth/react'
 import { toast } from 'react-toastify'
 import { zodResolver } from '@hookform/resolvers/zod'
 import { z } from 'zod'
+import { PencilIcon } from '@heroicons/react/24/solid'
```
vnugent: we're deprecating `@heroicons` library and are moving to Phosphor icons (sorry this is not documented anywhere).
Can you pick one from https://phosphoricons.com/?q=%22pencil%22?

vnugent:
Thanks for this PR!  A question about workflow: I uploaded a profile photo then uploaded another one.  It's working as expected.  What should we do with the previous profile photos? They seem unreachable/ aren't showing up on the gallery.

Would it be simpler (coding wise) to let the user pick from an existing photo?

vnugent (changes requested):

mikeschen:
> Thanks for this PR! A question about workflow: I uploaded a profile photo then uploaded another one. It's working as expected. What should we do with the previous profile photos? They seem unreachable/ aren't showing up on the gallery.
> 
> Would it be simpler (coding wise) to let the user pick from an existing photo?

So when a user uploads and then deletes a photo, that photo isn't actually deleted. If you go to gallery, delete a photo, then go to URL of deleted photo it is still there.
I originally had a user upload a profile pic, then it appears in gallery as well as profile but thought it was a weird UX (what happens to profile, when a user deletes the photo out of their gallery?)
Let me check how MP, Instagram and FB handle profiles....

mikeschen:
Looking at Instagram and MP (and Trailforks too) all have upload a new photo, I think it would be tricky to choose existing photo from gallery. (also of note, Instagram has remove photo, which leaves blank avatar).
None except Facebook has option to choose from existing.
But tell me what you think about uploading Profile, having it appear in Gallery and what to do with delete.
Or keep as is.

src/components/users/account/UpdateProfileForm.tsx
mikeschen: Gotcha No problem

vnugent:
We can keep the upload to profile as you implemented.  We want to be able to delete photos.  The user gallery has a caching bug otherwise it should delete photos from Google storage.

vnugent:
how about 
1.  Keep the upload logic as you implemented
2. Show profile photos in the gallery
3. If profile photo(s) were deleted, show a [face](https://phosphoricons.com/?q=%22face%22) or [user](https://phosphoricons.com/?q=%22user%22) icon

As an enhancement (TBD) we can add a flag to the photo upload API, to flag a profile photo?  This way we can filter how to display them on the frontend.  Thoughts?

mikeschen:
> how about
> 
> 1. Keep the upload logic as you implemented
> 2. Show profile photos in the gallery
> 3. If profile photo(s) were deleted, show a [face](https://phosphoricons.com/?q=%22face%22) or [user](https://phosphoricons.com/?q=%22user%22) icon
> 
> As an enhancement (TBD) we can add a flag to the photo upload API, to flag a profile photo? This way we can filter how to display them on the frontend. Thoughts?

That works for me 👍. I took a look at Strava too, they upload from files, and have delete also (although no choose from Strava gallery.) We can add flag to photo upload for profile in future.
Code is updated.

vnugent:
The reason photo delete doesn't work because it's broken #1263

PR #1259, opened 2025-01-23
fix: use Lead Style from MP tick import instead of Style

glassbead0:
---
name: Pull request
about: Create a pull request
title: 'correctly import attemptStyle from MP tick import'
labels: ''
assignees: ''

---

- [ ] refactor
- [ ] feature
- [X] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description
When importing ticks from Mountain Project, the `attemptStyle` field was being populated with the `Style` field from MP. It should use the `Lead Style` field instead.

### Related Issues

Issue #
The issue was created in openbeta-graphql, but the code is in open-tacos.

https://github.com/OpenBeta/openbeta-graphql/issues/424

### What this PR achieves
When converting a MPTick to a Tick object, set `attemptStyle` on Tick to be the value from `Lead Style` on MPTick (if present).

### Notes
I don't see an obvious way to write tests for this. everything in __tests__/ticks.ts mocks out Tick objects after the conversion from MPTick has already happened, so a new test would probably have to be more of an end-to-end tests.

Below  screenshot shows imported ticks locally, and the attemptType fields being populated with correct values.
Screenshot 2025-01-23 at 3 44 06 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,5 +80,8 @@ content
 *.pem
 .vscode/launch.json
 
+# for anyone using asdf
+.tool-versions
+
 .lighthouseci
 .idea/
--- a/src/app/api/user/bulkImportTicks/route.ts
+++ b/src/app/api/user/bulkImportTicks/route.ts
@@ -13,8 +13,8 @@ export interface Tick {
   notes: string
   climbId: string
   userId: string | undefined
-  style: string | null
-  attemptType: string | null
+  style: string | undefined
+  attemptType: string | undefined
   dateClimbed: Date
   grade: string
   source: string
@@ -61,7 +61,7 @@ interface MPTick {
 async function getMPTicks (profileUrl: string): Promise<MPTick[]> {
   const mpClient: AxiosInstance = axios.create({
     baseURL: 'https://www.mountainproject.com/user',
-    timeout: 60000
+    timeout: 180000
   })
   const res = await mpClient.get(`${profileUrl}/tick-export`)
   if (res.status === 200) {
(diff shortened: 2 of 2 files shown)


Vichy97:
Should we also coerce imported fields from MP into the set of possible values defined in https://github.com/OpenBeta/openbeta-graphql/pull/444? That could also be a separate PR, but it is necessary in order to make the change to allowable types

Vichy97 (approved):

glassbead0:
Update: I allowed null for both `style` and `attemptType`. The test pass, but i'm not sure if there's anywhere else in the codebase that needs updating to allow null values for these 2 fields.

vnugent:
Testing the PR locally (connecting to staging api, not local api server) I'm getting `ticks validation failed: attemptType: Path `attemptType` is required.`.  The error went away when I added 'N/A' back.

---

Screenshot 2025-01-26 at 11 30 29 AM

vnugent:
can you also increase the timeout to 180000 (3 mins)?
https://github.com/OpenBeta/open-tacos/pull/1259/files#diff-cc45e63ea0bcf9c96ed981bc5e05b06e060078e406cab1f16930ad29b93b9c38R76

glassbead0:
The issue of null vs 'N/A'...etc should be coordinated among these 3 PRs:
* (this one)
* https://github.com/OpenBeta/openbeta-graphql/pull/444/files#diff-d501761a02af29fcaa297f3f42d8f2f28006685a14bd07173f2e41c3f2e22542R81
* https://github.com/OpenBeta/open-tacos/pull/1260/files

Probably best discussed with mouth words if we meet this week.

PR #411, opened 2022-07-06
Link to blog

vnugent:

PR #1260, opened 2025-01-24
feat: Restrict attemptType based on style selected for ticks

glassbead0:
---
name: Pull request
about: Create a pull request
title: 'Restrict tick Attempt Type'
labels: ''
assignees: ''

---

- [X] refactor
- [X] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

[Issue #411](https://github.com/OpenBeta/openbeta-graphql/issues/411)

### What this PR achieves

* add the "Tick the Route" button to the newer climbs page (it was previously only on the old page)
* dynamically populate available attemptType based on chosen style
* display attemptType in TickCard when viewing listed ticks
* add subtle info icon with link to Wikipedia's Glossary of Climbing Terms
* tinker with z values because the breadcrumb nav bar was overlapping the tick dialog
* increase notes size slightly to accomodate for longer list of attempt styles in the dropdown.

### Screenshots, recordings

Screenshot 2025-01-24 at 5 06 16 PM

### Notes
this is closely tied to https://github.com/OpenBeta/openbeta-graphql/pull/444. and is essentially the front-end version of the same restrictions.

One thing i could use help with:
Style and attemptType can be empty, but right now thats getting sent as a string with a space. I'd like them to be nullable, but was running into issues with that.

### DIFF
--- a/src/components/ui/Tooltip.tsx
+++ b/src/components/ui/Tooltip.tsx
@@ -1,30 +1,22 @@
 import * as Popover from '@radix-ui/react-popover'
-import { ReactNode, useState } from 'react'
+import { ReactNode } from 'react'
 
 interface Props {
   content: string | ReactNode
   enabled?: boolean
   children: JSX.Element | JSX.Element [] | null
   className?: string
-  trigger?: 'click' | 'hover'
 }
 
 /**
  * A Tooltip that activates on mouse click or touch.
  * @param enabled false to disable tooltip but still render the trigger element
  * @param children Trigger element
- * @param trigger 'click' to activate on click, 'hover' to activate on hover. defaults to click
  */
-export default function Tooltip ({ content, enabled = true, className = '', children, trigger = 'click' }: Props): JSX.Element {
-  const [open, setOpen] = useState(false)
-  const handleMouseEnter = (): void => { if (trigger === 'hover') { setOpen(true) } }
-  const handleMouseLeave = (): void => { if (trigger === 'hover') { setOpen(false) } }
-
+export default function Tooltip ({ content, enabled = true, className = '', children }: Props): JSX.Element {
   return (
-    <Popover.Root open={trigger === 'hover' ? open : undefined} onOpenChange={setOpen}>
-      <Popover.Trigger className={className} onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseLeave}>
-        {children}
-      </Popover.Trigger>
+    <Popover.Root>
+      <Popover.Trigger className={className}>{children}</Popover.Trigger>
       {enabled &&
         <Content>
           {content}
--- a/src/components/ui/Tooltip.tsx
+++ b/src/components/ui/Tooltip.tsx
@@ -1,22 +1,30 @@
 import * as Popover from '@radix-ui/react-popover'
-import { ReactNode } from 'react'
+import { ReactNode, useState } from 'react'
 
 interface Props {
   content: string | ReactNode
   enabled?: boolean
   children: JSX.Element | JSX.Element [] | null
   className?: string
+  trigger?: 'click' | 'hover'
 }
 
 /**
  * A Tooltip that activates on mouse click or touch.
  * @param enabled false to disable tooltip but still render the trigger element
  * @param children Trigger element
+ * @param trigger 'click' to activate on click, 'hover' to activate on hover. defaults to click
  */
-export default function Tooltip ({ content, enabled = true, className = '', children }: Props): JSX.Element {
+export default function Tooltip ({ content, enabled = true, className = '', children, trigger = 'click' }: Props): JSX.Element {
+  const [open, setOpen] = useState(false)
+  const handleMouseEnter = (): void => { if (trigger === 'hover') { setOpen(true) } }
+  const handleMouseLeave = (): void => { if (trigger === 'hover') { setOpen(false) } }
+
   return (
-    <Popover.Root>
-      <Popover.Trigger className={className}>{children}</Popover.Trigger>
+    <Popover.Root open={trigger === 'hover' ? open : undefined} onOpenChange={setOpen}>
+      <Popover.Trigger className={className} onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseLeave}>
+        {children}
+      </Popover.Trigger>
       {enabled &&
         <Content>
           {content}
--- a/src/components/users/TickButton.tsx
+++ b/src/components/users/TickButton.tsx
@@ -11,6 +11,7 @@ interface Props {
   areaId?: string
   name?: string
   grade?: string
+  climbType: object
 }
 
 const IsTicked: React.FC<any> = ({ loading, onClick }) => {
@@ -25,7 +26,7 @@ const IsTicked: React.FC<any> = ({ loading, onClick }) => {
   )
 }
 
-export default function TickButton ({ climbId, areaId, name, grade }: Props): JSX.Element | null {
+export default function TickButton ({ climbId, areaId, name, grade, climbType }: Props): JSX.Element | null {
   const [loading, setLoading] = useState(false)
   const [isTicked, setIsTicked] = useState<boolean>(false)
   const [viewTicks, setViewTicks] = useState<boolean>(false)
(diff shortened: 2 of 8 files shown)


src/components/users/TickCard.tsx
```diff
@@ -37,7 +38,10 @@ export default function TickCard ({ tickId, ticks, setTicks, dateClimbed, notes,
         <p className='text-sm text-gray-500 mb-0'>{notes}</p>
       </div>
       <div className='flex flex-row'>
-        <p className='text-sm text-gray-500 pr-5'>{style}</p>
+        <div className='flex flex-col'>
```
glassbead0, line 41: This shows the attemptType on the tick card: 
Screenshot 2025-01-25 at 6 30 44 PM

src/pages/climbs/[id].tsx
```diff
@@ -440,7 +440,7 @@ const ClimbData: React.FC<{
 
         {!editMode && (
           <div className='mt-8'>
-            <TickButton climbId={climbId} name={name} grade={gradesObj.toString()} />
+            <TickButton climbId={climbId} name={name} grade={gradesObj.toString()} climbType={{}} />
```
Comment (glassbead0) [inline src/pages/climbs/[id].tsx:443]: Maybe I should just remove this tick buttons, since this is the "old" page....

src/components/users/TicksModal.tsx
```diff
@@ -12,7 +12,7 @@ const TicksModal: React.FC<any> = ({ open, setOpen, setTicks, setOpenForm, climb
   }
   return (
     <Transition.Root show={open} as={Fragment}>
-      <Dialog as='div' className='relative z-10' initialFocus={cancelButtonRef} onClose={setOpen}>
```
glassbead0, line 15: These z-values were needed so the breadcrumb nav bar didn't obscure the Tick modal.

src/components/users/TickForm.tsx
```diff
@@ -177,15 +235,23 @@ export default function TickForm ({ open, setOpen, setTicks, ticks, isTicked, cl
...
                     />
                   </div>
-                  <ComboBox options={styles} value={style} onChange={setStyle} label='Style' />
-                  <ComboBox options={attemptTypes} value={attemptType} onChange={setAttemptType} label='Attempt Type' />
+                  <ComboBox options={styles} value={style} onChange={handleStyleChange} label='Style' />
+                  <div className='flex items-center'>
+                    <label htmlFor='attemptType' className='block text-sm font-medium text-gray-700'>
+                      Attempt Type
+                    </label>
+                    <a href={climbingGlossaryLink} target='_blank' rel='noopener noreferrer' className='ml-2 text-gray-500 hover:text-gray-700' title='climbing terminology'>
```
glassbead0: The little info button is a link to Climbing Glossary. 
Screenshot 2025-01-25 at 6 33 55 PM

vnugent:
Does this PR supersede https://github.com/OpenBeta/open-tacos/pull/1259?  As commented in #1259 the backend doesn't like empty string for attemptType.  Maybe use 'N/A'?

src/components/users/TickForm.tsx
```diff
@@ -7,6 +7,7 @@ import { graphqlClient } from '../../js/graphql/Client'
 import { MUTATION_ADD_TICK } from '../../js/graphql/gql/fragments'
 import ComboBox from '../ui/ComboBox'
 import * as Yup from 'yup'
+import { InformationCircleIcon } from '@heroicons/react/20/solid'
```
vnugent: We're deprecating `@heroicons`.  Please use https://phosphoricons.com/

Example: https://github.com/OpenBeta/open-tacos/blob/develop/src/components/media/PhotoUploadButtons.tsx#L1

src/components/users/TickForm.tsx
vnugent: We have a Tooltip component you can use.  

Code https://github.com/OpenBeta/open-tacos/blob/develop/src/components/edit/form/AreaDesignationRadioGroup.tsx#L39-L41

---
Screenshot 2025-01-26 at 12 31 07 PM

src/pages/climbs/[id].tsx
Comment (vnugent) [reply src/pages/climbs/[id].tsx:443]: you can leave the button there until we delete the page

vnugent (changes requested):

glassbead0:
> Does this PR supersede #1259? As commented in #1259 the backend doesn't like empty string for attemptType. Maybe use 'N/A'?

It does not supersede it. #1259 deals with MP imports, and this one is all for the OB interface for adding ticks. I can certainly combine them, as they are related. I tend to like to combine related work into single PRs, but I also appreciate the value of separating things more granularly.

src/components/users/TickButton.tsx
```diff
@@ -1,3 +1,4 @@
+'use client'
```
glassbead0, line 1: I was getting errors locally telling me to add this here. "You're importing a component that needs useState. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default. Learn more: https://nextjs.org/docs/getting-started/react-essentials"

I don't know if thats a local-only issue for me though.

src/components/users/TickForm.tsx
```diff
@@ -19,35 +21,76 @@ const TickSchema = Yup.object().shape({
...
+
+  let styles: Array<{ id: number, name: string }> = []
+
+  if (leadable) { styles.push(...allStyles.filter(style => ['Lead', 'Follow'].includes(style.name))) }
+  if (boulderable) { styles.push(...allStyles.filter(style => ['Boulder'].includes(style.name))) }
+  if (topropeable) { styles.push(...allStyles.filter(style => ['TR'].includes(style.name))) }
+  if (aidable) { styles.push(...allStyles.filter(style => ['Aid'].includes(style.name))) }
+  if (soloable) { styles.push(...allStyles.filter(style => ['Solo'].includes(style.name))) }
+  if (styles.length === 0) { styles = [...allStyles] } // If a climb doesn't have a type, anything goes
+  styles.push({ id: 10, name: '\u00A0' })
```
glassbead0, line 77: FYI the '\u00A0' is the html encoded space character. this is to get the dropdown to show a blank option (and look nice). Down below, It handles this case and sends `undefined` to the backend, which gets handles correctly in this pr https://github.com/OpenBeta/openbeta-graphql/pull/444

src/components/ui/Tooltip.tsx
```diff
@@ -1,22 +1,30 @@
 import * as Popover from '@radix-ui/react-popover'
-import { ReactNode } from 'react'
+import { ReactNode, useState } from 'react'
 
 interface Props {
   content: string | ReactNode
   enabled?: boolean
   children: JSX.Element | JSX.Element [] | null
   className?: string
+  trigger?: 'click' | 'hover'
```
vnugent: I believe we default to click because hover isn't usable on web mobile

src/components/users/TickButton.tsx
vnugent: We do want to address that error.  Generally, we want to have as many server-side components as possible and add "use client" directive as escape hatch.  Maybe don't introduce "hover" behavior as it only works on desktop.

glassbead0: I'm happy to remove the hover functionality, but I don't think thats related to the 'use client' error.

Prior to my code changes, the TickButton relied on `useState` to determine if a route has been ticked by a user, so it knows whether to display `Tick This Route` or `View Ticks` on the button.

I'm not sure how to rewrite that functionality without useState, but i'm not at all an expert on Next.js.

I do see several other components that 'use client', so i'm not sure what the best solution is here.

src/components/ui/Tooltip.tsx
glassbead0: i did set default to click below on line 16, but regardless you're right, hover won't work on mobile. that said, my change here doesn't change the behavior of any other tooltip, only this specific one (which i wanted as a hover, because when you click it, it opens a link in a new tab).

I know this is an unnecessary tooltip at all though, and 'm totally happy to remove it if you think thats better.

vnugent: Let's not introduce a new behavior for Tooltip, and keep it consistent with existing UX (onClick show a tip, include a clickable external link if needed)

tooltip

vnugent (changes requested):

vnugent:
@all-contributors add @glassbead0 for code and ideas

Issue #1262, opened 2025-01-26
DWS, Snow, and Ice climbs display as 'Disciplines not set' in the UI, and Don't show grades


glassbead0:
## Steps to Reproduce
1. Find or create a route that has only 'DWS' as a discipline (repeat for Snow and Ice)
2. Navigate to the area page that contains the route
3. The DWS route in the list shows 'Disciplines Not Set` and does not display the grade
3. click the route
4. The route page also does not show a discipline or grade, and in the list of routes, no Discipline or grade is displayed

## Screenshots

![Image](https://github.com/user-attachments/assets/38876ed2-227e-4645-8aba-7c80fa2976b3)

![Image](https://github.com/user-attachments/assets/beb23f82-8d1b-4894-82af-0cc37544416d)

![Image](https://github.com/user-attachments/assets/6603589c-b0b0-492a-be63-0d33df7d954e)

## Expected Behavior
The route should show it's appropriate discipline (DWS, Ice or Snow) in these 3 places:
1) the area page, list of routes
2) the route page, list of routes
3) the route page, under the name of the route

## Browser & version
Chrome: 131.0.6778.265 

## Operating system
Example:  MacOS 15.2

PR #1282, opened 2025-02-19
fix: support all discipline types on climb and area pages

jkiernan12:
---
name: Pull request for Issue #1262 
about: Add ice, dws, and snow disciplines 
---

- [ ] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description
Ice, snow and DWS are missing in several places. This PR remedies that and consolidates the listing of some of these fields into a GraphQL fragment and a util function

### Related Issues

Issue #1262 

### What this PR achieves
Add missing climbing disciplines to areas and climb pages
- Create fragment for climbing disciplines and add to queries
- Add util to convert `ClimbDisciplineRecord` to simple array of `ClimbDiscipline[]`
- Remove now unused `disciplineArrayToObj` function

### Screenshots, recordings
Screenshot 2025-02-18 at 12 06 30 PM
Screenshot 2025-02-18 at 12 07 35 PM

### DIFF
--- a/src/app/(default)/climb/[[...slug]]/components/ClimbData.tsx
+++ b/src/app/(default)/climb/[[...slug]]/components/ClimbData.tsx
@@ -6,7 +6,7 @@ import RouteTypeChips from '@/components/ui/RouteTypeChips'
 import { ArticleLastUpdate } from '@/components/edit/ArticleLastUpdate'
 import { ClimbType, AreaType } from '@/js/types'
 import Grade from '@/js/grades/Grade'
-import { removeTypenameFromDisciplines } from '@/js/utils'
+import { removeTypenameFromDisciplines, getDisciplineList } from '@/js/utils'
 
 export const ClimbData: React.FC<ClimbType & Pick<AreaType, 'gradeContext'> & { isBoulder: boolean }> = (props) => {
   const { id, name, type, safety, length, grades, fa: legacyFA, authorMetadata, gradeContext, isBoulder } = props
@@ -28,7 +28,7 @@ export const ClimbData: React.FC<ClimbType & Pick<AreaType, 'gradeContext'> & { 
           {gradeStr != null && (
             <RouteGradeChip gradeStr={gradeStr} safety={safety} />
           )}
-          <RouteTypeChips type={type} />
+          <RouteTypeChips types={getDisciplineList(type)} />
         </div>
 
         {length !== -1 && (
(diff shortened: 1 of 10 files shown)


glassbead0 (changes requested):
@jkiernan12 for the most part this is awesome! So happy you pulled the list of types into a common location!!

When I check out your branch locally, I'm still seeing a couple issues, which you can see in the screenshot.
1.  On all routes (regardless of type), i'm seeing `__TYPENAME` displayed next to the route type - though I see in your screenshot that you don’t see that, which is interesting. I’ll investigate this a little and see if it’s a local issue for me.

2. On routes that have ONLY DWS, Ice, or Snow, I’m still not seeing the grade displayed on the route page, or in the list of routes, even though I set a grade when creating the route. When I create a route that has multiple types (like DWS and Sport), then I DO see the grade.

~I think the second issue is likely related to the sandbag library, and it’s possible there are some required code changes over there as well.~ 
Screenshot 2025-03-06 at 3 58 24 PM

Are you seeing the same issues as me? 

~It may get quite complicated pretty fast over in the sandbag library, as determining what kind of grade to assign to a "snow/trad" route is non-trivial, and eg: Ice/TR is also tricky.... TR routes typically have a YDS grade, but Ice routes have their own thing (WI? I'm not an ice climber)~

~At the very least, I think DeepWaterSolo should be straightforward - it should be a YDS grade (which the UI allows me to enter when I create a DWS route, but it just doesn't display the grade).~


I found the Grade issue: https://github.com/OpenBeta/open-tacos/blob/develop/src/js/grades/Grade.ts#L127 I think the logic here is incomplete - it doesn't account for all the route types. I'm not 100% what the logic there should be, but I tinkered around a little to at least account for all types. see my code snipped below.

src/app/(default)/climb/[[...slug]]/components/ClimbData.tsx
```diff
@@ -28,7 +28,7 @@ export const ClimbData: React.FC<ClimbType & Pick<AreaType, 'gradeContext'> & {
           {gradeStr != null && (
             <RouteGradeChip gradeStr={gradeStr} safety={safety} />
           )}
-          <RouteTypeChips type={type} />
+          <RouteTypeChips types={getDisciplineList(type)} />
```
Comment (glassbead0) [inline src/app/(default)/climb/[[...slug]]/components/ClimbData.tsx]: For me, `<RouteTypeChips types={getDisciplineList(sanitizedDisciplines)} />` worked to remove the `__TYPENAME` from displaying (although it gave me warnings about not passing the correct data type to the getDisciplineList function).

glassbead0:
I tinkered with the Grade.ts code to allow displaying grades for all route types:   The Ice/WI isn't quite right... i think there's some weirdness around that that might need tracking down, but it's close.

~Here's a commit to see the changes: https://github.com/glassbead0/open-tacos/commit/bf3315efac7f66602b134b81188310544b3e290f~

~@jkiernan12 would you be willing to cherry-pick that commit, and test it out locally?~

~You'll probably also need this small backend change: 
https://github.com/OpenBeta/openbeta-graphql/pull/455/files~

The backend change has been merged, and I pushed the new commit to this branch. this PR should be ready to go. waiting for build.
Screenshot 2025-03-18 at 7 52 22 PM

glassbead0 (approved):

PR #1314, opened 2025-03-23
fix: remove duplicate html tags

vishalpatel011:
## Issue
When loading pages from the app router (e.g. https://stg.openbeta.io/pulse), there are duplicate `<html>` tags in the source, causing hydration issues during client-side rendering.

## Changes Made
1. Removed `<html>` and `<body>` tags from `src/app/(default)/layout.tsx`
2. Kept the HTML structure only in the root layout (`src/app/layout.tsx`)
3. Renamed the component from `RootLayout` to `DefaultLayout` for clarity

## Testing
- [x] Verified that pages load without duplicate HTML tags
- [x] Confirmed that all functionality (auth, header, footer) works as expected
- [x] Tested on Chrome browser

### DIFF
--- a/src/app/(default)/layout.tsx
+++ b/src/app/(default)/layout.tsx
@@ -22,24 +22,20 @@ export const metadata: Metadata = {
 /**
  * Global layout for the site
  */
-export default function RootLayout ({
+export default function DefaultLayout ({
   children
 }: {
   children: React.ReactNode
 }): any {
   return (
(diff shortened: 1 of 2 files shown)


mikeschen:
Nice work!

mikeschen (approved):
Looks good on my end

vishalpatel011:
Thank you! Feels great to make my first open source contribution 😄

PR #1320, opened 2025-04-07
update error msg for user name change

mikeschen:
---
name: Pull request
about: Create a pull request
title: ''
labels: ''
assignees: ''

---

- [ ] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description
Go to Edit Profile
Go to Username
Change username
Click Save
Go to change name again
See error msg (must wait 14 days)
### Related Issues

Issue #460 (in OpenBeta repo)

### DIFF
--- a/src/components/users/account/UsernameChangeForm.tsx
+++ b/src/components/users/account/UsernameChangeForm.tsx
@@ -96,7 +96,7 @@ export const UsernameChangeForm: React.FC = () => {
     } catch (e) {
       console.error(e)
       reset()
-      toast.error('Unexpected error.  Please try again.')
+      toast.error(e instanceof Error ? e.message : 'Unexpected error.  Please try again.')
     }
   }
 


src/components/users/account/UsernameChangeForm.tsx
```diff
@@ -96,7 +96,7 @@ export const UsernameChangeForm: React.FC = () => {
     } catch (e) {
       console.error(e)
       reset()
-      toast.error('Unexpected error.  Please try again.')
+      toast.error(e instanceof Error ? e.message : 'Unexpected error.  Please try again.')
```
glassbead0, line 99: So does this correctly show an error message of `Must wait 14 days to change username`, instead of the unhelpful `Unexpected Error` message?

I see [this issue](https://github.com/OpenBeta/openbeta-graphql/issues/460) on the backend, but it seems to me the 14 day wait is intentional, and this error message is the only thing that needs fixing. Should we close that backend issue, or am I missing something?

One thing that doesn't make sense to me is that I went to the deployment for this PR, and I was able to change the username several times in a row. I dont' see how this code change bypasses the 14 day  limit 🤔 

cc @mikeschen

mikeschen: This is the FE component of the ticket.
I just put out PR for BE here: https://github.com/OpenBeta/openbeta-graphql/pull/463

vnugent:
Backend deployed

Issue #1346, opened 2025-07-14
Add test area somewhere on website landing page


emilyg1709:
Noticed someone making a test area in Uganda instead of in the designated "test area" in the USA that serves this purpose. 

Might be a good improvement to direct new users to use the test area to learn the site, so we don't end up with "test" boulders in a bunch of random locations.

icebelly29:
I would like to work on this issue. @mikeschen can you assign this to me?

icebelly29:
Hi @mikeschen 

I came across two possible area URLs:

- `/area/18c5dd5c-8186-50b6-8a60-ae2948c548d1/whole-hazelnuts`
- `/area/fedcd484-cf47-5263-ab2a-52dc27c738e0/utah-upload-test`

Could you please confirm which one is the official Test Area we should be linking to from the homepage?

mikeschen:
We would want to delete those areas. We would want a "secret" url, one that is not linked to anything, that we can direct devs to. something like: /area/18c5dd5c-8186-50b6-8a60-ae2948c548d1/test-area

mikeschen:
> the designated "test area" in the USA that serves this purpose.
@emilyg1709 where is this test area ?

emilyg1709:
@mikeschen @icebelly29 this is the link to the test area where people can play with editing modes and such. Might be helpful if there were a button somewhere on the home page (maybe near latest contributions or something) that says something like "Learn to use the site here" and people can go to the test area where they aren't adding bad beta on existing climbs or places. 

I only added this because I noticed somebody adding areas called test in random places and imo that's not really something we want happening. 

https://openbeta.io/area/18c5dd5c-8186-50b6-8a60-ae2948c548d1/test-area

mikeschen:
<img width="307" height="257" alt="Image" src="https://github.com/user-attachments/assets/b6d1d9b3-e5eb-4924-bbdd-e0f390d2b6e9" />

Maybe this , I posted to the Discord channel for opinions.

mikeschen:
Ah ok so we can swap out the banner when a user is logged in:
When not logged in, show this:
```
'use client'
import Cookies from 'js-cookie'
import { useState, useEffect } from 'react'
import { useSession } from 'next-auth/react'
import { SuppressButton } from './SuppressButton'

const STORAGE_KEY = 'suppress-main-banner'

export interface AppAlertProps {
  message: JSX.Element
}

/**
[...]

PR #1353, opened 2025-07-22
feat: add Test Area banner on homepage to guide new users

icebelly29:
---
name: Pull request
about: Create a pull request
title: Add Test Area banner to homepage to guide new users
labels: ''
assignees: ''

---

- [ ] refactor
- [x] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

## Description
This PR adds a "Test Area" banner just below the homepage hero section to help guide new users.

Fixes Issue #1346 

### What this PR achieves

This PR adds a visually distinct yet unobtrusive banner below `<LandingHero />` that links to the official Test Area, using consistent styling to encourage proper testing practices among new contributors.

### Screenshots, recordings
Screenshot 2025-07-22 015012


mikeschen (commented):
Looks good, yarn.lock has some changes that snuck in if you can revert those.
 ~~Also lets make "Share Your Route Climbning knowledge banner the same size.
We can add a "DOn't show again" button too.~~

mikeschen (commented):
I think we are on the right track if we combine the Share Your Route Climbing knowledge banner with the Test Area banner.

icebelly29:
> Looks good, yarn.lock has some changes that snuck in if you can revert those. 

just to confirm: did you mean to fully exclude the lockfile changes from this PR (which is done), or were there intentional updates you’d like to keep instead?

mikeschen (commented):
It looks like yarn.lock got deleted. We need that file since it is the single source of truth for packages that are installed.

mikeschen (commented):
Thinking a little more about this, I realised we have `AppAlert.tsx` component, we should utlize it correctly. We just need to add a prop called `cookieStorageKey` 
Then remove useSession from the component and use it in the parent.
In the parent `LandingHero.tsx` we will have two `AppAlert.tsx` components.
ONe will be the exising "Share your climbing route knowledge!",(shown if a user is not logged in)
the other will be the new "Try the test area" (shown when a user is logged in)
Let me know if this makes sense.

icebelly29:
>  the other will be the new "Try the test area" (shown when a user is logged in) Let me know if this makes sense.

I’m working on the Test Area banner logic and would like to verify its behavior for logged-in users. Would it be possible to get test account credentials or a way to log in locally to simulate an authenticated user?

mikeschen:
Can you try signing up for account on openbeta.io , then using that login?

PR #1379, opened 2025-09-22
fix: Decode HTML entities in climb and area names

### LINKED ISSUE #1242: ampersands in route names get mangled
## Steps to Reproduce

1. Create a route with "&" in its name 

## Screenshots

https://openbeta.io/climb/2e82d08c-6cb4-42c8-be39-fafd230ec432/climb-with-an-ampersand-in-name-andamp
Screenshot_20241212_215750

## Expected Behavior

The ampersand should be rendered as "&".

## Current Behavior

The ampersand is rendered as `&amp;`.

## Browser & version

Firefox 133.0.2

## Operating system

Android

AnshulJagotra:
Fixes #1242

Problem Description
Climb and area names containing ampersands (&) or other HTML entities were rendered incorrectly in the frontend.
Currently, these entities (&amp;, &lt;, &gt;, etc.) were displayed literally instead of being decoded, causing route names and climb/area names to appear incorrectly.

Solution

Added a safeDecode utility function in src/js/utils.ts to safely decode common HTML entities.

Updated the following components to use safeDecode before rendering names:

ClimbData.tsx

Area page (page.tsx)

ClimbListForm.tsx

Ensures proper display in both client-side rendering (CSR) and server-side rendering (SSR).

Idempotent solution: safe even if backend resolves encoding in the future.

Testing

Verified that climb and area names with & or other HTML entities display correctly in the UI.

Verified SSR pages render correctly without crashing.

User-facing changes

Climb and area names now display correctly (& instead of &amp;) on all pages.

Backwards-incompatible changes

None. This change only affects frontend rendering logic; existing data and backend APIs remain unchanged.

Optional / Suggestion

If the backend (REST or GraphQL API) is encoding HTML entities before sending names, it may be preferable to send plain text directly.

Frontend will continue to work safely even if the backend implements this change in the future.

### DIFF
--- a/src/app/(default)/area/[[...slug]]/page.tsx
+++ b/src/app/(default)/area/[[...slug]]/page.tsx
@@ -9,7 +9,7 @@ import { getAreaRSC } from '@/js/graphql/getAreaRSC'
 import { StickyHeaderContainer } from '@/app/(default)/components/ui/StickyHeaderContainer'
 import { AreaCrumbs } from '@/components/breadcrumbs/AreaCrumbs'
 import { ArticleLastUpdate } from '@/components/edit/ArticleLastUpdate'
-import { getMapHref, getFriendlySlug, getAreaPageFriendlyUrl, sanitizeName, parseUuidAsFirstParam } from '@/js/utils'
+import { getMapHref, getFriendlySlug, getAreaPageFriendlyUrl, sanitizeName, parseUuidAsFirstParam, safeDecode } from '@/js/utils'
 import { LazyAreaMap } from '@/components/maps/AreaMap'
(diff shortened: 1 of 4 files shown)


mikeschen (commented):
I see that these functions are created, but where are they called? @AnshulJagotra

AnshulJagotra:
Thanks @mikeschen, you were right — the safeDecode function was created but not used in the components.
I’ve now updated ClimbData.tsx, Area page.tsx, and ClimbListForm.tsx to call safeDecode when rendering names.

mikeschen:
Request changes not completed. Closing for now.

Issue #1184, opened 2024-10-28
Add a search bar to the /maps route


clintonlunn:
Currently there is not a way to search for different areas/climbs in the `/maps` route. This means that the user needs to zoom and pan to the area/climb they want to find. 

We should add a search bar that will allow the user to search by area/climb (like the main page search bar), then zoom to the the area and open the relevant popup

In addition, we should probably add a header bar to hold the home button, profile button, and search bar. The "Crags/boulders" and "Boundaries" layer selector might be nice to continue to have floating under that header bar. 

<img width="874" alt="image" src="https://github.com/user-attachments/assets/8b940588-d22a-4d2f-a656-56c985ead219">

<img width="354" alt="image" src="https://github.com/user-attachments/assets/6e195020-731d-45d9-9f29-05ccf99fb1d6">

**How important is this to you** (Please pick one)
- Important

lexicallearner:
hey can i work on this issue ?

PR #1384, opened 2025-10-12
added the search bar as mentioned in the issue no. #1184

GaneshGupta3:
---
name: Pull request
about: Create a pull request
title: ''
labels: ''
assignees: ''

---

- [x] refactor
- [x] feature
- [ ] bug fix
- [ ] documentation
- [ ] optimization
- [ ] other

### Related Issues

Issue #

### DIFF
--- a/src/app/(maps)/components/FullScreenMap.tsx
+++ b/src/app/(maps)/components/FullScreenMap.tsx
@@ -7,6 +7,7 @@ import { MapLayerMouseEvent } from 'maplibre-gl'
 import { useUrlParams } from '@/js/hooks/useUrlParams'
 import { lineString, Position } from '@turf/helpers'
 import lineToPolygon from '@turf/line-to-polygon'
+import { Search, X } from 'lucide-react'
 
 export const FullScreenMap: React.FC = () => {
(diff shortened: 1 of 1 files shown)


clintonlunn:
I do not think this would be in line with [nominatim's usage policy](https://operations.osmfoundation.org/policies/nominatim/). Portions I suspect will be in violation are: 

- No heavy uses (an absolute maximum of 1 request per second).
- Provide a valid HTTP Referer or User-Agent identifying the application (stock User-Agents as set by http libraries will not do).
- Auto-complete search - the debounces search that is happening here is an implementation of autocomplete.

This would not be an issue if there was an openbeta hosted nominatim instance. 

@mikeschen

mikeschen:
CLOSING due to:
@clintonlunn  I do not think this would be in line with [nominatim's usage policy](https://operations.osmfoundation.org/policies/nominatim/). Portions I suspect will be in violation are:

No heavy uses (an absolute maximum of 1 request per second).
Provide a valid HTTP Referer or User-Agent identifying the application (stock User-Agents as set by http libraries will not do).
Auto-complete search - the debounces search that is happening here is an implementation of autocomplete.
This would not be an issue if there was an openbeta hosted nominatim instance.

PR #1433, opened 2026-01-04
fix(image): replace deprecated Next.js Image props

clintonlunn:
onLoadingComplete was deprecated in Next.js 14.2

https://nextjs.org/docs/app/api-reference/components/image#deprecated-props


mikeschen (approved):
Looks good!

Issue #1452, opened 2026-03-27
AppAlert causes layout shift for first-time visitors


devjhex:
## Bug Description
I noticed that `AppAlert` initializes with `useState(false)` which causes a layout shift for first-time visitors. Since `useEffect` runs after the first render, the alert pops in after the page has already painted, pushing content down.

## Current Behaviour

- Page renders with alert hidden
- `useEffect` runs and checks for a cookie
- If no cookie is found, alert appears and pushes content down

## Expected Behaviour
Alert renders visible from the start for users with no cookie, so there is no content shift on initial load.

## Notes
The alert sits right above the content and images on the landing page. First-time visitors naturally start looking at the page as it loads and then the alert suddenly pops in and pushes everything down  not a great first impression.

## Before Layout Shit

<img width="1504" height="504" alt="Image" src="https://github.com/user-attachments/assets/f35b65ce-a042-4165-9038-29a0a5f9e0d6" />

## After Layout Shift

<img width="1505" height="612" alt="Image" src="https://github.com/user-attachments/assets/53e66d8f-bc79-410c-9a88-dbeef1dec10c" />

PR #1453, opened 2026-03-27
fix(AppAlert): prevent layout shift on initial render for new users

devjhex:
---
name: Pull request
about: Create a pull request
title: 'Fix layout shift for main alert component'
labels: 'bug fix'
assignees: ''

---

- [ ] refactor
- [ ] feature
- [x] bug fix
- [ ] documentation
- [x] optimization
- [ ] other

### What this PR achieves
AppAlert was initializing with ```useState(false)``` which caused a layout shift for every first-time visitor. Since useEffect runs after the first render, the alert would always pop in after the page painted, pushing content down.

Changing to ```useState(true)``` means the alert renders visible from the start — no shift for new users. The only tradeoff is a brief flash for returning users who already dismissed it (cookie exists), but this is barely noticeable especially on fast devices and affects a much smaller group.

### Related Issues

Issue #1452

## Illustration
shift-illustration

### Notes

- First-time visitors — the most critical audience — now get a completely clean experience with zero layout shift regardless of their network conditions. The only tradeoff is a brief flash for returning users who already dismissed the alert, which is practically invisible on fast devices and only affects a group that has already seen the site before.

### DIFF
--- a/src/components/broadcast/AppAlert.tsx
+++ b/src/components/broadcast/AppAlert.tsx
@@ -16,7 +16,7 @@ export const AppAlert: React.FC<AppAlertProps> = ({
   message,
   cookieStorageKey
 }) => {
-  const [showAlert, setShowAlert] = useState(false)
+  const [showAlert, setShowAlert] = useState(true)
 
   useEffect(() => {
     const suppressed = Cookies.get(cookieStorageKey)


mikeschen:
We would prefer not to have brief flash for returning users who already dismissed the alert. Layout shift is minimal for new users.

Issue #1466, opened 2026-08-27
Can't edit profile unless adding data to all fields.


CaptainZidgel:
Create a new account. Go to edit profile. There are three fields, implied optional. `Save changes` btn is disabled as nothing has been changed.
Add information to one of the fields. Btn remains disabled but should be enabled.
Add information to each field (including a _valid_ URL for a personal website).
`Save changes` btn is now enabled.

I should be able to update a display name without adding a bio and url.

From a cursory look at the code, I believe the issue is in the Zod resolver schema not returning isValid properly, https://github.com/OpenBeta/open-tacos/blob/develop/src/components/users/account/UpdateProfileForm.tsx#L45C45-L45C52 , but I couldn't find any information in the Zod docs about what could be causing this.

PR #1467, opened 2026-09-01
fix: allow editing profile with optional fields

Tanvinaik09:
Title:
fix: handle undefined optional profile fields

[x] bug fix

Related Issues:
Issue #1466

What this PR achieves:
Fixes the profile edit form validation when optional fields are empty.

Previously, empty profile fields could contain null or undefined values, causing the form to remain invalid even when only one field was updated.

This change normalizes displayName, bio, and website to empty strings when resetting the form, ensuring optional empty fields are handled correctly.

Notes:
The change is limited to normalizing optional profile fields during form reset.


Tanvinaik09:
i have fixed the issue, can you please review it?

